PR #1, opened 2016-12-29
MessageContext::get_message

stasm:
This aligns the `MessageContext` public API to that of its JS equivalent. `format` takes a Value object which must be first acquired via `get_message`.

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -8,10 +8,13 @@ fn main() {
     ctx.add_messages("key1 = Value 1");
     ctx.add_messages("key2 = Value 2");
 
-    let msg = ctx.get_message("key1").unwrap();
-
-    match ctx.format(&msg) {
-        Ok(v) => println!("{}", v),
-        Err(err) => println!("Couldn't retrieve: {:?}", err),
+    if let Some(msg) = ctx.get_message("key1") {
+        match ctx.format(&msg) {
+            Ok(v) => println!("Formatted value: {}", v),
+            Err(err) => println!("Formatting error: {:?}", err),
+        }
+    } else {
+        println!("Missing message: {}", "key1");
     }
+
 }
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -8,10 +8,10 @@ fn main() {
     ctx.add_messages("key1 = Value 1");
     ctx.add_messages("key2 = Value 2");
 
-    let id = String::from("key1");
+    let msg = ctx.get_message("key1").unwrap();
 
-    match ctx.format(&id) {
-        Ok(v) => println!("Message for key {} - {}", &id, v),
-        Err(err) => println!("Couldn't retrieve the entry {:?}: {:?}", &id, err),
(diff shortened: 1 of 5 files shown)


src/context/mod.rs
```diff
@@ -30,15 +30,14 @@ impl MessageContext {
...
-        match self.entries.get(id) {
-            Some(ref value) => {
-                match resolve(self, value) {
-                    Ok(msg) => Ok(msg),
-                    Err(_) => Err(ContextError::Generic),
-                }
-            }
-            None => Err(ContextError::Generic),
+    pub fn get_message(&self, id: &str) -> Option<&ast::Value> {
+        self.entries.get(id)
```
stasm: I understand that right now `MessageContext` uses the runtime AST. I wonder if that's needed. Could we do without this special version of the AST in the Rust implementation?

examples/simple.rs
```diff
@@ -8,10 +8,10 @@ fn main() {
     ctx.add_messages("key1 = Value 1");
     ctx.add_messages("key2 = Value 2");
 
-    let id = String::from("key1");
+    let msg = ctx.get_message("key1").unwrap();
```
zbraniecki: `unwrap` is usually a sign of a trouble to come. It's useful for mocks, but we should avoid them in code as much as possible.

In this case, since `get_message` returns an Option,  I think there's a value in matching it, and printing an error message in case it returns false (instead of panicking).

src/context/mod.rs
```diff
@@ -12,42 +12,36 @@ pub enum ContextError {
...
             match entry {
-                ast::Entry::Message { id, value, traits } => {
-                    self.entries.insert(id.clone(),
-                                        ast::Entry::Message {
-                                            id: id,
-                                            value: value,
-                                            traits: traits,
-                                        });
+                ast::Entry::Message(msg @ ast::Message { .. }) => {
+                    self.messages.insert(msg.id.clone(), msg);
```
zbraniecki: +1! :)

src/syntax/runtime/parser.rs
```diff
@@ -90,15 +90,16 @@ pub fn parse(source: &str) -> Result<ast::Resource> {
 
     let mut entries = vec![];
 
-    let entry = get_entity(&mut ps)?;
+    let message = get_entity(&mut ps)?;
+    let entry = ast::Entry::Message(message);
 
     entries.push(entry);
```
zbraniecki: Can you inline the entry value into `push` please? No need for a separate variable that is used only once in the next line.

examples/simple.rs
stasm: Got it, thanks for explaining.

src/syntax/runtime/parser.rs
stasm: Sure. Should I keep `let message = …` as it is to make the `?` more explicit?

zbraniecki: yeah, let's leave the `let message = ...` :)

PR #2, opened 2016-12-30
Serde

zbraniecki:
Let's see if that will fly.

I want to use serde to serialize FTL to json. I actually want to serialize to at least two types of JSON - pure AST serialization and a particular one for the entries-json so that the library can be a ready-to-use replacement for l20n.js.

I also want fluent not to require serde, since I like the idea of keeping fluent-rs lean and for anyone who doesn't need to dump it to JSON it's not a dependency.
Maybe one day all the parsing code will end up in separate crate and we won't need serde in libfluent.

So, I decided to start by using the conditional compilation features.
This allows us to have two optional features - `json` and `entries-json`, which mingle with `parser.rs` and add serde/json to libfluent. Depending on which feature you ask for, we'll compile with the `syntax::json` or `syntax::runtime::json` - the latter being for entries, the former for pure JSON serialization.

I also moved AST to `syntax:ast` and `syntax:parser` which means we now only have one parser/ast. I think that it's a good change.

Beyond changes to parser, and adding the two `json.rs` I don't think there are any major changes, but I'd like you to look at it and verify that you like the direction.

wrt. parser.rs, I can see us ending up with separate binaries for each scenario, but for now, I think that `cfg` options should be enough.

Merge at will.


stasm:
What is your take on what the MVP for this repo is?  Is it a standalone implementation of Fluent in Rust, or is it a Rust parser to be used with l20n.js's Resolver?  Perhaps we'd like to have two crates (and possibly repos) for these two goals?

zbraniecki:
Good point.

I thought of it as a standalone implementation of Fluent in Rust, that's why I wanted to keep the entries.json serialization as an option.

I can try to move that out into sth like `fluent-ast-json-rs` and `fluent-l20njs-json-rs` (names tbd)

stasm:
Something along these lines might be good. Perhaps it would be also helpful to define the goal this l20n.js-compatible parser would serve in the bigger scheme of things.

zbraniecki:
> Perhaps it would be also helpful to define the goal this l20n.js-compatible parser would serve in the bigger scheme of things.

At this stage I guess I'm in the exploration land of trying to figure out what is possible in Rust and what is not.

But I believe that my goals for the two things are:

1) Have a implementation independent serialization format that can be parsed by rust, and read by python or js. JSON fits this needs perfectly.

2) Have a drop-in rust replacement for the parser so that we can test using it in Gecko instead of l20n.js' parser to test performance win.

Do you share the view that having those two would be a value?

If the answer is yes, I believe it's easier to explore the configuration setup that will allow us to achieve that, even if we won't write the whole thing now. Deciding if we need "features" in fluent-rs, or can we go with a separate crate is easier now when the crate is small.

zbraniecki:
Ok, it's not super easy because I cannot just add traits to structs/enums that are not from my crate.
That means that it's impossible to do sth like `impl Serialize for fluent::syntax::ast::Resource`.

What I have to do is duplicate the whole AST and add `From` for each `ast::Node` to local node and then add serialize/deserialize traits for that.

Luckily, it hasn't been that much of a hassle, and since performance is not the crucial part now, I think I'm ok continuing with two separate crates for serializing/deserializing fluent to json.

https://github.com/zbraniecki/fluent-json-rs

If we ever end up needing to get json into fluent-rs, we can, and if we'll decide that we need an efficient parser that parses to entries-json it should be fairly easy to create it out of fluent-rs and fluent-js-entries-rs.

For now, it should be enough to let me create a drop-in replacement for l20n.js Parser for Gecko without having to mingle with fluent-rs crate! :)

Closing.

stasm:
> 1. Have a implementation independent serialization format that can be parsed by rust, and read by python or js. JSON fits this needs perfectly.

To achieve this goal, would you like to standardize the AST? I was under the impression that we chose to standardize the syntax precisely so that we don't have to standardize the AST. As a result, implementations may design their own ASTs which work well with the language they're written in.

That said, I don't expect those implementation-specific ASTs to be much different from each other.  if need be, we can also have _full_ and _lean_ AST versions, similar to what we did in l20n.js.

> 2. Have a drop-in rust replacement for the parser so that we can test using it in Gecko instead of l20n.js' parser to test performance win.

Yes, that would be interesting to know. This is what zbraniecki/fluent-json-rs aims to be, right? Maybe we don't have to implement the whole parser to be able to test this with l20n.js's resolver?

zbraniecki:
> To achieve this goal, would you like to standardize the AST? 

No, I'm ok saying that this is fluent-rs AST and we only guarantee that `fluent-json-rs` can serialize and deserialize it.

> As a result, implementations may design their own ASTs which work well with the language they're written in.

Agree.

> Yes, that would be interesting to know. This is what zbraniecki/fluent-json-rs aims to be, right? Maybe we don't have to implement the whole parser to be able to test this with l20n.js's resolver?

Yeah, we don't, but I'd like to get it to pass fixtures entries.json and workload-low.ftl which I use for benchmarks. I have the code in l20n.rs, so it's all about porting now.

The real hassle is that I have to write pretty daunting From and Into that basically does manual conversion, and that there will be a performance hit coming from those conversions, but I think it's fine for now. If we'll end up having a production need for performance, we can create another crate that will merge fluent-rs and fluent-js-entries-rs into one super-fast parser :)

PR #3, opened 2016-12-30
Implement EntityReference

stasm:
This PR adds support for evaluating `Patterns` with `TextElements` and
`MessageReferences`.  The parser doesn't support it yet.

I simplified error-handling to returning `Option<String>` for now.  Consequently
I removed unused error enums.  I'd like to move to using the `Result` type in the
future, keeping in mind that a single formatting operation can result in more
than one error.

I also simplified the AST for now.  Almost everything is an `Expression` now and
a `Message`'s value isn't restricted to `Patterns`.  It made it easier to write
the eval method.  The Parser should still enforce the correct syntax.


zbraniecki:
> I simplified error-handling to returning Option<String> for now.

Why can't we use errors?

> I also simplified the AST for now. Almost everything is an Expression now and
a Message's value isn't restricted to Patterns. It made it easier to write
the eval method. The Parser should still enforce the correct syntax.

I'd like to discuss it more. I'd prefer not to go the route of turning everything into an Expression.

stasm:
> Why can't we use errors?

We can, but I think I'd prefer to design error-handling in a separate issue. I didn't feel comfortable making ad-hoc decisions about it right now. For instance, I'm not sure if in a `Result<T, E>`-typed return value, `E` should be a single error or a vector of errors (of more than one placeable fail).  Or, maybe it should be a `struct Fallback { value: String, errors: Vec<Error> }`?

> I'd prefer not to go the route of turning everything into an Expression.

My thinking here was that `TextElement` is an expression that needs to be evaluated but is not accepted by the grammar and the parser anywhere else than in a `Pattern`.  It shouldn't stop us from making it an `Expression` in the AST.

stasm:
> I'd prefer not to go the route of turning everything into an Expression.

Can you explain why?

zbraniecki:
> We can, but I think I'd prefer to design error-handling in a separate issue. I didn't feel comfortable making ad-hoc decisions about it right now. For instance, I'm not sure if in a Result<T, E>-typed return value, E should be a single error or a vector of errors (of more than one placeable fail). Or, maybe it should be a struct Fallback { value: String, errors: Vec<Error> }?

Making it return `ContextError::Generic` for now will work, you can always design it later.
But there's no value in returning `Option` over `Result`, except that you can't use the `try` semantics.

> My thinking here was that TextElement is an expression that needs to be evaluated but is not accepted by the grammar and the parser anywhere else than in a Pattern.

So, I think I'm not sure why you want to get every node in AST to be an Expression. I'm not saying that it's impossible, but it seems to me that it looses the AST value because you can end up with some variant of

```
struct Expression {
  field: Expression,
  field2: Expression
}

struct Expression2 {
  field: Vec<Expression>
}
```

in which case the whole AST is just a generic Map.

The way I wanted to see it is an AST that actually represents the nodes we have in the language and get to the point where the AST can be validated with the RBNF grammar. You seem to be moving away from that.

Parser will of course produce the right thing, that's not my concern. But with your approach I envision anything writing AST directly (like, l10n editors, merge tools etc.) to be able to easily design completely different ASTs that we won't be able to handle.
So, that's on the cost side.

I don't see anything on the value side. Can you elaborate on why would you want the `TextElement` to be an expression, instead of asking my why I would not want it to be one? :)

stasm:
> Making it return ContextError::Generic for now will work, you can always design it later.
But there's no value in returning Option over Result, except that you can't use the try semantics.

I implemented the simplest strategy for now: return `None` when there's no value to format, and return `Some(String)` when the value exists.  If evaluating the value produces errors (e.g. a `MessageReference` to an unknown message or to a message without a value), the missing pieces are filled in with `___`.  This allows the result to always be a string.

 I considered using the `Result` type. The trouble I had is to know when to return `ContextError::Generic`. When a single `MessageReference` doesn't eval cleanly, should we just return an error and be done with it? Or should we somehow collect all errors from all placeables and return both the salvaged string value of the pattern and the errors? That sounds like something I'd like to discuss in a separate issue.

> So, I think I'm not sure why you want to get every node in AST to be an Expression.

My definition of an `Expression` is: a node in the AST which evaluates to something. `Patterns`, `MessageReferences` or `CallExpressions` are all expressions. OTOH, traits, variants of `SelectExpressions` or key-value arguments in `CallExpressions` are not, but can be part of one.

I think we'll end up with something like (using the current vocabulary):

```rust
struct Message {
    id: String,
    value: Option<Expression>,
    traits: Option<Vec<Member>>,
}

struct Member {
    key: String,
    value: Expression,
    default: bool,
}

enum Expression {
    TextElement(String),
    MessageReference(String),
    ExternalArgument(String),
    SelectExpression {
        selector: Expression,
        variants: Vec<Member>
    },
    Pattern {
        elements: Vec<Expression>,
    },
    // etc.
}
```

Different expression types have different fields and the whole thing produces a tree. While the grammar doesn't allow for anything else than a `Pattern` to be a message's value, it doesn't matter in the AST: the value is something that needs to be evaluated, so it can simply be an `Expression`.

> The way I wanted to see it is an AST that actually represents the nodes we have in the language and get to the point where the AST can be validated with the RBNF grammar. You seem to be moving away from that.

Grammar-validation of an AST is hard. Perhaps you were thinking of CSTs, Concrete Syntax Trees? CSTs can be thought of as another representation of the grammar and can usually be automatically generated from it. ASTs are by definiton abstract and can diverge from the grammar. 

> But with your approach I envision anything writing AST directly (like, l10n editors, merge tools etc.) to be able to easily design completely different ASTs that we won't be able to handle.

Enforcing grammar rules via strong typing can only go so far. For instance, there's not way to make sure a `Member`'s key doesn't use forbidden characters. Strong typing would make it easier to avoid mistakes on compile time—but that's usually not when tools make mistakes. They modify the tree on runtime. I think a more sound (and long-term) solution would be to offer good serialization and validation libraries for tools.

> Can you elaborate on why would you want the TextElement to be an expression, instead of asking my why I would not want it to be one? :)

:)

Like I said in https://github.com/projectfluent/fluent-rs/pull/3#issuecomment-269840718, `TextElements` are evaluated as part of `Patterns` which makes them good candidates for being `Expressions`.  Since we're working towards an MVP, there's value in keeping things simple. It allows us to iterate quickly.

zbraniecki:
> That sounds like something I'd like to discuss in a separate issue.

Ok, I see your point. Agree.

> Since we're working towards an MVP, there's value in keeping things simple. It allows us to iterate quickly.

I moved the conversation out to issue #5 . Let's discuss is there and come back here when we're ready.

PR #4, opened 2016-12-31
Add README

stasm:
I keep talking about the MVP. Let's define what it means :) @zbraniecki how about this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "fluent"
-version = "0.1.0"
-authors = ["Zibi Braniecki <gandalf@mozilla.com>"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "fluent"
-version = "0.1.0"
-authors = ["Zibi Braniecki <gandalf@mozilla.com>"]
+version = "0.0.1"
+authors = [
+    "Zibi Braniecki <gandalf@mozilla.com>",
+    "Staś Małolepszy <stas@mozilla.com>"
+]
```
zbraniecki: +1 :)

Issue #5, opened 2016-12-31
Design an AST for fluent-rs


zbraniecki:
Let's start an issue to design how we want our AST to look like.

zbraniecki:
@stasm proposal is:
```
struct Message {
    id: String,
    value: Option<Expression>,
    traits: Option<Vec<Member>>,
}

struct Member {
    key: String,
    value: Expression,
    default: bool,
}

enum Expression {
    TextElement(String),
    MessageReference(String),
    ExternalArgument(String),
    SelectExpression {
        selector: Expression,
        variants: Vec<Member>
    },
    Pattern {
        elements: Vec<Expression>,
    },
    // etc.
}
```

zbraniecki:
My proposal is:

```rust
struct Resource {
  body: Vec<Entry>,
  comment: Option<Comment>
}

struct Message {
  id: Identifier,
  value: Option<Pattern>,
  traits: Option<Vec<Member>>,
  comment: Option<Comment>
}

struct Comment {
  body: String
}

struct Section {
  key: Key,
  body: Vec<Entry>,
  comment: Option<Comment>
}

enum Entry {
  Message(Message),
  Comment(Comment),
  Section(Section),
  JunkEntry(JunkEntry)
}

struct Identifier {
  name: String,
}

struct Key {
  name: String
}

struct Number {
  value: String
}

struct Keyword {
  ns: Option<String>,
  name: Key
}

struct Pattern {
  elements: Vec<PatternElement>,
  quoted: bool
}

enum PatternElement {
  Text(String),
  Placeable {
    expressions: Vec<Expression>
  }
}

struct MemberKey {
  Keyword(Keyword),
  Number(Number)
}

struct Member {
  key: MemberKey,
  value: Pattern,
  default: bool,
}

pub enum Expression {
  ExternalArgument {
    id: Identifier
  },
  EntityReference {
    id: Identifier
  },
  FunctionReference {
    id: Identifier
  },
  Number(Number),
  CallExpression {
    callee: Box<Expression>,
    args: Vec<Expression>
  },
  SelectExpression {
    expression: Box<Expression>,
    variants: Vec<Member> 
  },
  KeyValueArgument {
    name: String,
    value: Box<Expression>
  },
  MemberExpression {
    object: Box<Expression>,
    keyword: MemberKey,
  }
}

struct JunkEntry {
  content: String,
}
```

zbraniecki:
Several notes about my approach:

[...]

PR #6, opened 2017-01-01
Parse simple placeables

zbraniecki:


zbraniecki:
Pending #5.

src/syntax/ast.rs
```diff
@@ -1,40 +1,67 @@
...
 pub struct Pattern {
     pub elements: Vec<PatternElement>,
+    pub quoted: bool,
 }
 
 #[derive(Debug, PartialEq)]
 pub enum PatternElement {
     Text(String),
-    Placeable(Vec<Expression>),
+    Placeable { expression: Vec<Expression> },
```
stasm: Maybe call it `expressions` (plural) until we make the syntax changes discussed in tools-l10n?

PR #12, opened 2017-02-01
Merge spec 0.2 updates

zbraniecki:


zbraniecki:
@stasm - can you take a look? The parser code is less important (since I'll keep working on it and we'll get more strengthening for 0.3), but I'd like to get your approval on AST and test changes.

stasm:
The idea behind introducing the ASDL description of the AST was to help document what the data structure should look like. I see a few discrepancies between the code in ast.rs and https://github.com/projectfluent/syntax/blob/61ef8fc961ad7b2df4a8873b09e8165dae27612a/fluent.asdl. I'll comment inline.

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
-}
-
-#[derive(Debug, PartialEq)]
-pub struct Section {
-    pub key: Key,
-    pub body: Vec<Entry>,
+pub struct Message {
+    pub id: Identifier,
+    pub value: Option<Pattern>,
+    pub attributes: Option<Vec<Attribute>>,
```
stasm: +1 for `attributes`, I'll change the ASDL.

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
-    Comment(Comment),
     Section(Section),
+    Comment(Comment),
     Junk(JunkEntry),
 }
 
 #[derive(Debug, PartialEq)]
-pub struct Message {
+pub struct Pattern {
+    pub elements: Vec<PatternElement>,
```
stasm: Do we still need `PatternElement`? Can you add `Text(String)` to `Expression` and make `elements` here be a `Vec<Expression>`?

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
+    AttributeExpression {
+        id: Identifier,
+        attr: Identifier
+    },
+    KeyValueArgument {
+        id: Identifier,
+        val: ArgValue
+    },
+    Number(Number),
+    Placeable(Box<Expression>),
```
stasm: Good catch here, I'll need to add `Expression` here in the ASDL as well.

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
+}
+
+#[derive(Debug, PartialEq)]
+pub enum SelectorExpression {
+    Pattern(Pattern),
+    MessageReference(String),
+    ExternalArgument(String),
+    CallExpression {
+        callee: Builtin,
+        args: Vec<SelectorExpression>,
```
stasm: Shouldn't this just be `Vec<Expression>`? I understand that we don't want to allow bare `SelectExpressions` as arguments here, and they require to be wrapped in `{ }`, but I think this should be encoded in the EBNF rather than the AST. There's no ambiguity in the AST and the braces don't have to be part of it.

OTOH, I see how this helps serialization. If the AST contains the information about the braces used for grouping, the process of serialization doesn't have to check the parent of a `SelectExpression` to decide if it should be wrapped in braces.

stasm: If we leave this as-is, would you agree to the following renames?

- `Expression` → `PlaceableExpression`,
- `SelectorExpression` → `Expression`

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
+    },
+    VariantExpression {
+        id: Identifier,
+        key: VariantKey
```
stasm: The ASDL uses `msg` and `key` here. I like `id` as well, so I'll change it.

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
+    },
+    AttributeExpression {
+        id: Identifier,
+        attr: Identifier
```
stasm: The ASDL uses `msg` and `id` here. I like your idea of using `id` instead of `msg`.  I would suggest `name` instead of `attr`. `attr` suggest it's the whole attribute object, just like `var` would suggest it's the whole variant and not a variant key in `VariantExpression`.

src/syntax/ast.rs
```diff
@@ -5,101 +5,122 @@ pub struct Resource {
...
+    },
+    KeyValueArgument {
```
stasm: We abuse the word _key_ in the AST. Would you agree to rename this to `NamedArgument { name, val }`?

stasm:
Inspired by this PR, I landed https://github.com/projectfluent/syntax/commit/9453242714d6bac38ec18b8db8289108dd54d28f which makes the ASDL closely mimic fluent-rs's AST. I went ahead and made a few renames which I summarized in the commit message.

I also submitted https://github.com/projectfluent/syntax/pull/26 in which I propose to simplify the AST. Let's discuss there and come back here for the implementation part.

zbraniecki:
Thanks! I'll update the AST/Parser once the pr lands

zbraniecki:
Actually, I updated the PR already :)

src/syntax/ast.rs
```diff
@@ -5,101 +5,108 @@ pub struct Resource {
...
+    pub body: String,
 }
 
 #[derive(Debug, PartialEq)]
-pub enum MemberKey {
-    Keyword(Keyword),
-    Number(Number),
+pub struct Section {
+    pub key: Keyword,
+    pub body: Vec<Entry>,
```
stasm: Remove `body`?

src/syntax/ast.rs
```diff
@@ -5,101 +5,108 @@ pub struct Resource {
...
-        name: Identifier,
-        val: Box<Expression>,
-    },
-    Number(Number),
-    String(String),
-    Member {
-        obj: Box<Expression>,
-        key: MemberKey,
-    },
+pub struct Builtin {
```
stasm: Maybe rename to `Function`? That's how we refer to them in the docs. Also, the ASDL defines the `name` field.

src/syntax/ast.rs
```diff
@@ -5,101 +5,108 @@ pub struct Resource {
...
+    pub quoted: bool,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Expression {
+    Pattern(Pattern),
+    String(String),
+    Number(Number),
+    MessageReference(String),
+    ExternalArgument(String),
```
stasm: I don't know how closely you'd like to keep the Rust implementation with the ASDL, so I'll let you decide here. The ASDL specifies the `id` field on `MessageReference` and `ExternalArgument`:

    MessageReference(iden id)
    ExternalArgument(iden id)

which in Rust would map to:

    MessageReference { id: Identifier },
    ExternalArgument { id: Identifier },

(which was in the ast.rs before this PR).

src/syntax/ast.rs
```diff
@@ -5,101 +5,108 @@ pub struct Resource {
...
+
+#[derive(Debug, PartialEq)]
+pub struct Identifier {
     pub name: String,
 }
 
 #[derive(Debug, PartialEq)]
-pub struct Number {
+pub struct Keyword {
     pub value: String,
```
stasm: This used to be `name` which is why it's name in ASDL. I don't care either way, should I change it?

zbraniecki: I'll change it

PR #14, opened 2017-02-28
Update to spec 0.3

zbraniecki:


src/syntax/ast.rs
```diff
@@ -104,11 +105,6 @@ pub struct Number {
...
 #[derive(Debug, PartialEq)]
-pub struct Function {
-    pub name: String,
-}
-
-#[derive(Debug, PartialEq)]
-pub struct JunkEntry {
-    pub content: String,
+pub struct Comment {
+    pub body: String,
```
stasm: Rename `body` to `content` here.

src/context.rs
```diff
@@ -16,21 +16,33 @@ impl MessageContext {
         let res = parse(source).unwrap_or_else(|x| x.0);
 
         for entry in res.body {
+            let id = match entry {
+                ast::Entry::Message{ref id, ..} => {
+                    id.name.clone()
+                }
+                _ => unimplemented!(),
```
stasm: IIUC, this will panic on Comments or Sections. Is this intended?  The Context doesn't use either of these Entry types.

tests/fixtures/parser/ftl/05-variants-errors.ftl
```diff
@@ -1,28 +1,37 @@
...
+}
 
-key5 =
+key5 = {
  *fw] Foo
+}
 
-key6 =
+key6 = {
   [ a] A
```
stasm: Many of these examples will error simply because there's no default variant in the select expression. Consider adding a second valid one, e.g. `*[ok] Valid`.

src/syntax/ast.rs
```diff
@@ -28,7 +31,6 @@ pub struct Pattern {
 
 #[derive(Debug, PartialEq)]
 pub enum Expression {
-    Pattern(Pattern),
     String(String),
     Number(Number),
```
stasm: Please rename to `StringExpression` and `NumberExpression` as per ASDL. Also, shouldn't we define a separate `StringExpression` struct with a `value` field instead of using Rust's native string?

src/syntax/ast.rs
```diff
@@ -28,7 +31,6 @@ pub struct Pattern {
 
 #[derive(Debug, PartialEq)]
 pub enum Expression {
-    Pattern(Pattern),
```
stasm: Here, I think we need an `Element` or `PatternElement` enum with `TextElement` and `Expression` in it.

zbraniecki: I'm wondering if instead we should have:
```
pub struct Pattern {
  pub elements: Vec<PatternElement>
}

pub enum PatternElement {
  TextElement(String),
  Expression(Expression)
}
```

Because adding `PatternElement` into `Expression seems to be useless as those elements can be anywhere except of inside Pattern.elements.

src/syntax/ast.rs
zbraniecki: Why would we need `StringExpression` struct? What field would it hold except of a string? Where else would it be used?

src/syntax/ast.rs
stasm: Yes, sorry, that's exactly what I meant. Sorry, I didn't mean it should be added into the `Expression` enum. Bad placement of my comment.

src/syntax/ast.rs
stasm: I was simply going by the ASDL and noticed that it was missing.  If you think it's not needed and that `StringExpression { value }` nodes can be added to JSON during the serialization, that's totally fine too.

PR #15, opened 2017-07-04
Resolver

zbraniecki:
I think this is ready for review. It adds basic MessageReference and ExternalArgument handling and I'd like to get the outline right for future work :)

### DIFF
--- a/examples/external_arguments.rs
+++ b/examples/external_arguments.rs
@@ -1,6 +1,7 @@
 extern crate fluent;
 
 use fluent::context::MessageContext;
+use fluent::context::FluentArgument;
 use std::collections::HashMap;
 
 fn main() {
@@ -9,9 +10,10 @@ fn main() {
 
     ctx.add_messages("hello-world = Hello { $name }");
     ctx.add_messages("ref = The previous message says { hello-world }");
+    ctx.add_messages("unread-emails = You have { $emailCount } unread emails");
 
     let mut args = HashMap::new();
-    args.insert(String::from("name"), String::from("John"));
+    args.insert(String::from("name"), FluentArgument::from("John"));
 
     match ctx.get_message("hello-world")
               .and_then(|msg| ctx.format(&msg, Some(&args))) {
@@ -20,6 +22,15 @@ fn main() {
     }
 
     match ctx.get_message("ref")
+              .and_then(|msg| ctx.format(&msg, Some(&args))) {
+        Some(value) => println!("{}", value),
+        None => println!("None"),
+    }
+
+    let mut args = HashMap::new();
+    args.insert(String::from("emailCount"), FluentArgument::from(5));
+
+    match ctx.get_message("unread-emails")
               .and_then(|msg| ctx.format(&msg, Some(&args))) {
         Some(value) => println!("{}", value),
         None => println!("None"),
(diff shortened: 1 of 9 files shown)


stasm:
> It adds basic MessageReference and ExternalArgument handling […]

The PR moves a bunch of code to a new file so it's not easy for me to answer this myself: did you change anything in the implementation of `MessageReference`?

examples/external_arguments.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    ctx.add_messages("hello-world = Hello { $name }");
+    ctx.add_messages("ref = The previous message says { hello-world }");
+    ctx.add_messages("unread-emails = You have { $emailCount } unread emails");
+
+    let mut args = HashMap::new();
+    args.insert(String::from("name"), FluentArgument::from("John"));
+
+    match ctx.get_message("hello-world")
+              .and_then(|msg| ctx.format(&msg, Some(&args))) {
```
stasm: Nit: The indentation looks off here.

examples/external_arguments.rs
```diff
@@ -0,0 +1,38 @@
...
+fn main() {
+    let locales = vec![String::from("pl")];
+    let mut ctx = MessageContext::new(locales);
+
+    ctx.add_messages("hello-world = Hello { $name }");
+    ctx.add_messages("ref = The previous message says { hello-world }");
+    ctx.add_messages("unread-emails = You have { $emailCount } unread emails");
+
+    let mut args = HashMap::new();
+    args.insert(String::from("name"), FluentArgument::from("John"));
```
stasm: Is there a way to use `&str` as keys of the `args` map? I'm not sure how `args.get` would then look them up, but having to use `String::from` looks a bit unwieldy.

src/types.rs
```diff
@@ -0,0 +1,12 @@
+pub enum FluentType {
+    FluentString(String),
+    FluentNone,
+    FluentNumber,
+}
+
+pub fn value_of(t: FluentType) -> String {
```
stasm: Could we implement this as a custom trait on each of the `FluentTypes`?

examples/external_arguments.rs
zbraniecki: That's what `cargo fmt` did :)

src/types.rs
zbraniecki: You can implement traits on enum variants yet - https://internals.rust-lang.org/t/impl-trait-for-enum-variant/4131/3

zbraniecki:
> The PR moves a bunch of code to a new file so it's not easy for me to answer this myself: did you change anything in the implementation of MessageReference?

I tried not to. It's still in `eval_expr` and formats a message or returns `___`. The only diff is that it now passes the arguments as well.

zbraniecki:
@stasm - I added a very basic test for ExternalArgument. I can add more but wanted to check with you if this is the kind of test you'd like to see.

zbraniecki:
> Yeah, something like this is great. Maybe test a number and some other type which fails in FluentArgument::from. And referencing a message which uses external arguments.

I added two of those tests. I can't add a test for a type that fails in ::from because... it's a compilation error (yay Rust!) :)

I believe we can improve on the ::value_of and ::from in future commits. Gonna merge this.

PR #18, opened 2017-09-29
Resolver 0.1

stasm:
This depends on #17.

- [x] `AttributeExpression` in placeables and selectors,
- [x] `VariantExpression` in placeables and selectors,
- [x] `MessageRefence` in selectors (using tags),
- [x] `MessageContext::format` is able to format attributes,
- [x] A simple mock-plural rule is used in `FluentValue::matches`,
- [x] `ast::Number` now uses `f32` under the hood.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use super::syntax::ast;
 use super::syntax::parse;
 use super::types::FluentValue;
-use super::resolve::resolve;
+use super::resolve::{Env, ResolveValue};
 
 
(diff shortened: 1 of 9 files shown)


stasm:
I rebased this on top of master which now includes #17.

stasm:
> I think we'll want to revisit how we use &str later on (I suspect Env should own its string keys), but we can do this as a separate cleanup for the whole crate later :)

Sounds good. I plan to add protection against cyclic references which in the JS implementation uses the `Env.dirty` set, so perhaps I can refactor the `&str` usage there too.

PR #29, opened 2017-10-13
Change formatting in README.md example

nvzqz:
Closure parameters usually follow immediately after the opening parenthesis for the called function.

Chained methods also (subjectively) look nicer when aligned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,8 @@ use fluent::MessageContext;
 let ctx = MessageContext::new(&["en-US"]);
 ctx.add_messages("hello-world = Hello, world!");
 
-let value = ctx.get_message("hello-world").and_then(
-    |message| ctx.format(message, None),
-);
+let value = ctx.get_message("hello-world")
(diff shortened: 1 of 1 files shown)


zbraniecki:
Thank you! Merged

stasm:
The original formatting came from running `cargo fmt`. I find `rustfmt` is very helpful because it completely offloads the decision-making involved in formatting the syntax.

FWIW, this is what the README's example would look like formatted with the stable branch of `rustfmt`:

```rust
extern crate fluent;

use fluent::MessageContext;

fn main() {
    let mut ctx = MessageContext::new(&["en-US"]);
    ctx.add_messages("hello-world = Hello, world!");

    let value = ctx.get_message("hello-world").and_then(|message| {
        ctx.format(message, None)
    });

    assert_eq!(value, Some("Hello, world!".to_string()));
}
```

stasm:
I submitted #33 to enable running `cargo fmt` in CI.

PR #33, opened 2017-10-16
Run rustfmt on Travis

stasm:
`rustfmt` makes code formatting a non-issue. Let's run in in CI as a reminder to autoformat code before merging.


stasm:
The second commit here changes `rustfmt` to `rustfmt-nightly`. I'm not sure we should make this change just yet.  It would be good to first understand how stable `rustfmt-nightly`'s formatting style is at this point.

PR #35, opened 2017-10-19
Travis: Export LD_LIBRARY_PATH for libsyntax.so

stasm:


zbraniecki:
force-installing cargo moves the job time from 3min to 20 min :( I would love us to find a different solution that doesn't require rebuilding travis-cargo on each job.

stasm:
Well, it's better to have jobs that run 20 minutes and complete than to have it fail. It's pretty bad that #34 which doesn't touch any code is red.

stasm:
I would even recommend removing coveralls completely if it makes Travis run tests in a reliable way.

zbraniecki:
Yeah, I agree that it's better to lose coverage than to fight this :(

stasm:
It looks like the reason why it fails is that it can't find `libsyntax.so`. So it might be related to `rustfmt` and not `coveralls`.

stasm:
Yeah, removing `coveralls` completely didn't fix it.

stasm:
Looks like exporting `LD_LIBRARY_PATH` fixed it. Cool.

PR #37, opened 2017-12-13
Remove tags

zbraniecki:
First - remove tags. I'll re-enable the test once we enable private messages.

### DIFF
--- a/src/resolve.rs
+++ b/src/resolve.rs
@@ -99,28 +99,6 @@ impl ResolveValue for ast::Expression {
                 ref variants,
             } => select_default(variants).and_then(|variant| variant.value.to_value(env)),
             ast::Expression::SelectExpression {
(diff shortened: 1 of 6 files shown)


src/resolve.rs
```diff
@@ -99,25 +99,9 @@ impl ResolveValue for ast::Expression {
                 ref variants,
             } => select_default(variants).and_then(|variant| variant.value.to_value(env)),
             ast::Expression::SelectExpression {
-                expression: Some(box ast::Expression::MessageReference { ref id }),
+                expression: Some(box ast::Expression::MessageReference { .. }),
```
stasm: Is there a reason to still special-case `MessageReference` used as a selector? There's another `ast::Expression::SelectExpression` branch in this `match` right below and I'd think it's enough?

PR #38, opened 2017-12-15
Add private messages

zbraniecki:


src/syntax/errors/list.rs
```diff
@@ -69,17 +70,17 @@ pub fn get_error_desc(err: &ErrorKind) -> (&'static str, String, &'static str) {
...
+            "Keyword cannot end with a whitespace".to_owned(),
             "",
         ),
-        ErrorKind::ForbiddenCallee => (
+        ErrorKind::ForbiddenCaller => (
             "E0008",
-            "a callee has to be a simple identifier".to_owned(),
-            "",
+            "A caller has to be a simple identifier".to_owned(),
+            "Remove `-` from in front of the caller",
```
stasm: I think `The caller` might be more suitable here. Also, `Remove…` isn't a good advice, if a corresponding message doesn't exist.

src/syntax/errors/list.rs
```diff
@@ -88,9 +89,14 @@ pub fn get_error_desc(err: &ErrorKind) -> (&'static str, String, &'static str) {
...
         ),
         ErrorKind::MissingVariants => (
-            "E0010",
+            "E0011",
             "Expected at least one variant after \"->\".".to_owned(),
             "",
         ),
+        ErrorKind::ForbiddenAttributeExpression => (
+            "E0012",
+            "Private message attributes cannot be used as placeables.".to_owned(),
```
stasm: `Attributes of private messages…`

src/syntax/ftlstream.rs
```diff
@@ -205,17 +213,24 @@ where
...
-    fn take_id_start(&mut self) -> Result<char> {
+    fn take_id_start(&mut self, allow_private: bool) -> Result<char> {
         let closure = |x| match x {
-            'a'...'z' | 'A'...'Z' | '_' => true,
+            'a'...'z' | 'A'...'Z' => true,
+            '-' => allow_private,
             _ => false,
         };
 
         match self.take_char(closure) {
```
stasm: Consider `if let` here.

src/syntax/parser.rs
```diff
@@ -222,13 +222,13 @@ where
     Ok(attributes)
 }
 
-fn get_identifier<I>(ps: &mut ParserStream<I>) -> Result<ast::Identifier>
+fn get_identifier<I>(ps: &mut ParserStream<I>, allow_private: bool) -> Result<ast::Identifier>
```
stasm: For readability, I would very much like to see two convenience wrappers around `get_identifier`: `get_public_identifier` and `get_private_identifier`.

src/syntax/parser.rs
```diff
@@ -512,6 +512,12 @@ where
         } else {
             ps.reset_peek();
         }
+    } else {
+        if let ast::Expression::AttributeExpression { ref id, .. } = selector {
+            if id.name.starts_with('-') {
+                return error!(ErrorKind::ForbiddenAttributeExpression);
```
stasm: There should be a similar check in the branch which returns a new `SelectExpression` (or an early return before it) but in the other direction: only attributes of private messages can be selectors ([EBNF](https://github.com/projectfluent/fluent/blob/06d20408febd5c50da1c6b6e173568b403c07ba5/spec/fluent.ebnf#L74)).

src/syntax/parser.rs
```diff
@@ -642,20 +651,34 @@ where
 {
     if let Some(ch) = ps.current() {
         let exp = match ch {
-            '0'...'9' | '-' => ast::Expression::NumberExpression {
+            '0'...'9' => ast::Expression::NumberExpression {
                 value: get_number(ps)?,
             },
+            '-' => match ps.peek() {
```
stasm: Consider `if let` here.

src/syntax/parser.rs
```diff
@@ -538,6 +544,9 @@ where
                 Ok(ast::Expression::VariantExpression { id, key: key })
             }
             Some('(') => {
+                if id.name.starts_with('-') {
+                    return error!(ErrorKind::ForbiddenCaller);
```
stasm: The `ForbiddenCaller` check needs a bit of work, likely outside of this PR. The EBNF requires the callee to be [all uppercase](https://github.com/projectfluent/fluent/blob/06d20408febd5c50da1c6b6e173568b403c07ba5/spec/fluent.ebnf#L30).

src/syntax/parser.rs
stasm: We should also disallow `VariantExpressions` as selectors, as per the [EBNF](https://github.com/projectfluent/fluent/blob/06d20408febd5c50da1c6b6e173568b403c07ba5/spec/fluent.ebnf#L64).

src/syntax/errors/list.rs
```diff
@@ -69,17 +72,17 @@ pub fn get_error_desc(err: &ErrorKind) -> (&'static str, String, &'static str) {
         }
         ErrorKind::ForbiddenWhitespace => (
             "E0007",
-            "keyword cannot end with a whitespace".to_owned(),
+            "Keyword cannot end with a whitespace".to_owned(),
             "",
         ),
-        ErrorKind::ForbiddenCallee => (
+        ErrorKind::ForbiddenCaller => (
```
stasm: I think `Callee` was correct?

src/syntax/errors/list.rs
```diff
@@ -69,17 +72,17 @@ pub fn get_error_desc(err: &ErrorKind) -> (&'static str, String, &'static str) {
...
             "E0008",
-            "a callee has to be a simple identifier".to_owned(),
+            "The caller has to be a simple, upper-case, identifier".to_owned(),
             "",
         ),
         ErrorKind::ForbiddenKey => (
             "E0009",
-            "a key has to be a simple identifier".to_owned(),
+            "A key has to be a simple identifier".to_owned(),
```
stasm: Since you're changing it already: `The key...` or `Keys...` both sound better to me.

src/syntax/errors/list.rs
```diff
@@ -88,9 +91,24 @@ pub fn get_error_desc(err: &ErrorKind) -> (&'static str, String, &'static str) {
             "",
         ),
         ErrorKind::MissingVariants => (
-            "E0010",
+            "E0011",
             "Expected at least one variant after \"->\".".to_owned(),
             "",
         ),
+        ErrorKind::ForbiddenAttributeExpression => (
```
stasm: Perhaps rename this to `ForbiddenPrivateAttributeExpression` for consistency with `ForbiddenPublicAttributeExpression `?

src/syntax/ftlstream.rs
```diff
@@ -205,17 +213,25 @@ where
...
 
-        match self.take_char(closure) {
-            Some(ch) => Ok(ch),
-            None => error!(ErrorKind::ExpectedCharRange {
-                range: String::from("'a'...'z' | 'A'...'Z' | '_'"),
-            }),
+        if let Some(ch) = self.take_char(closure) {
+            Ok(ch)
+        } else {
+            if allow_private {
```
stasm: Nit: perhaps flatten this into an `else if`?

PR #39, opened 2017-12-18
0.5 comments

zbraniecki:
Built on top of the private msg PR. Need to add tests, but hopefully a good start :)


src/syntax/parser.rs.orig
```diff
@@ -0,0 +1,721 @@
+pub use super::errors::ParserError;
```
stasm: Please remove this file.

src/syntax/parser.rs
```diff
@@ -152,6 +143,11 @@ where
 {
     let id = get_private_identifier(ps)?;
 
+    if let Some(ast::Comment::Comment { .. }) = comment {
+    } else if comment.is_some() {
```
stasm: I don't understand this code. It looks like both `if let` and `else if` test if `comment` is a `Some`?

zbraniecki: heh, yeah, it's awkward. What I'm trying to test is if the variant of the Comment passed to get_message is actually `ast::Comment::Comment` and not GroupComment or ResourceComment.

If it's not, then I'm verifying that it's None.

Any ideas how to write it better? :)

stasm: Ahh, I see now, thanks. I might have wrongly suggested to use `if let` here then. Maybe a `match` like this?

```rust
match comment {
    Some(ast::Comment::GroupComment { .. }) => error!(),
    Some(ast::Comment::ResourceComment { .. }) => error!(),
    _ => ()
}
```

In the future, https://github.com/rust-lang/rfcs/issues/935 will help.

PR #40, opened 2017-12-21
Add block expressions

zbraniecki:
This is a bit larger diff than I would hope for, but it makes our parser more aligned with the ebnf. I have a high hope to get a more significant rewrite/cleanup of the parser as I make it copy-free, so I'm trying to balance out how much I refactor now.

This gets the 0.5 syntax to work.

### DIFF
--- a/tests/fixtures/parser/ftl/errors/09-block-expr.ftl
+++ b/tests/fixtures/parser/ftl/errors/09-block-expr.ftl
@@ -1,0 +1,7 @@
+key = Test { $foo ->
+  *[other] Value
+}
+
+key = { $foo ->
+  *[other] Value
+} Test
--- a/src/intl/mod.rs
+++ b/src/intl/mod.rs
@@ -481,7 +481,7 @@ fn get_plural_rule(loc: &str) -> Option<usize> {
         "zu" => 3,
         _ => return None,
     };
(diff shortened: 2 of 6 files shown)


zbraniecki:
ping @stasm ! :)

stasm:
I wish GitHub had a lace where it shows me all my review requests. Sorry I missed this, I'll review it this week.

tests/fixtures/parser/ftl/errors/09-block-expr.ftl
```diff
@@ -0,0 +1,7 @@
+key = Test { $foo ->
+  *[other] Value
+}
+
+key = { $foo ->
+  *[other] Value
+} Test
```
stasm: If I'm reading the code right, this second `key` is actually successfully parsed up to the newline after `  *[other] Value` and then an error is produced for the remaining `} Test`. I find this behavior surprising; I'd expect the whole message to be treated as Junk.

stasm:
I filed https://github.com/projectfluent/fluent/pull/81 to remove the block expression sugar from Fluent Syntax 0.5. Let's consider it for the [next milestone](https://github.com/projectfluent/fluent/milestone/5) of the Syntax and let's hold on with the changes in this PR.

PR #41, opened 2017-12-22
Fix travis

zbraniecki:


stasm:
Did this fix Travis by disabling the linter check?

PR #45, opened 2018-02-02
Support Fluent Syntax 0.5

zbraniecki:
Aligns the parser with python_fluent as much as possible. Lots of pythonisms that I'd like to work on after landing this.
Also, had to disable two tests due to the unfortunate messages/terms split. I'd like to brainstorm solutions outside of this patch as well :)

### DIFF
--- a/src/resolve.rs
+++ b/src/resolve.rs
@@ -34,6 +34,12 @@ impl ResolveValue for ast::Message {
     }
 }
 
+impl ResolveValue for ast::Term {
+    fn to_value(&self, env: &Env) -> Option<FluentValue> {
+        self.value.to_value(env)
+    }
+}
+
 impl ResolveValue for ast::Attribute {
     fn to_value(&self, env: &Env) -> Option<FluentValue> {
         self.value.to_value(env)
@@ -62,12 +68,6 @@ impl ResolveValue for ast::PatternElement {
     }
 }
 
-impl ResolveValue for ast::Placeable {
-    fn to_value(&self, env: &Env) -> Option<FluentValue> {
-        self.expression.to_value(env)
-    }
-}
-
 impl ResolveValue for ast::Number {
     fn to_value(&self, _env: &Env) -> Option<FluentValue> {
         f32::from_str(&self.value).ok().map(FluentValue::from)
@@ -87,10 +87,12 @@ impl ResolveValue for ast::Expression {
                 Some(FluentValue::from(value.clone()))
             }
             ast::Expression::NumberExpression { ref value } => value.to_value(env),
+            ast::Expression::MessageReference { ref id } if id.name.starts_with('-') => env.ctx
+                .get_term(&id.name)
+                .and_then(|term| term.to_value(env)),
             ast::Expression::MessageReference { ref id } => env.ctx
                 .get_message(&id.name)
-                .and_then(|message| message.value.as_ref())
-                .and_then(|pattern| pattern.to_value(env)),
+                .and_then(|message| message.to_value(env)),
             ast::Expression::ExternalArgument { ref id } => env.args
                 .and_then(|args| args.get(&id.name.as_ref()))
                 .cloned(),
@@ -127,10 +129,15 @@ impl ResolveValue for ast::Expression {
                 select_default(variants).and_then(|variant| variant.value.to_value(env))
             }
             ast::Expression::AttributeExpression { ref id, ref name } => {
-                let attributes = env.ctx
-                    .get_message(&id.name)
-                    .as_ref()
-                    .and_then(|message| message.attributes.as_ref());
+                let attributes = if id.name.starts_with('-') {
+                    env.ctx
+                        .get_term(&id.name)
+                        .and_then(|term| term.attributes.as_ref())
+                } else {
+                    env.ctx
+                        .get_message(&id.name)
+                        .and_then(|message| message.attributes.as_ref())
+                };
                 if let Some(attributes) = attributes {
                     for attribute in attributes {
                         if attribute.id.name == name.name {
@@ -140,27 +147,23 @@ impl ResolveValue for ast::Expression {
                 }
                 None
             }
-            ast::Expression::VariantExpression { ref id, ref key } => {
-                let message = env.ctx.get_message(&id.name);
-                let variants = message
-                    .as_ref()
-                    .and_then(|message| message.value.as_ref())
-                    .and_then(|pattern| {
-                        if pattern.elements.len() > 1 {
-                            return None;
-                        }
+            ast::Expression::VariantExpression { ref id, ref key } if id.name.starts_with('-') => {
+                let term = env.ctx.get_term(&id.name);
+                let variants = term.as_ref().and_then(|term| {
+                    if term.value.elements.len() > 1 {
+                        return None;
+                    }
 
-                        match pattern.elements.first() {
-                            Some(&ast::PatternElement::Placeable(ast::Placeable {
-                                expression:
-                                    ast::Expression::SelectExpression {
-                                        expression: None,
-                                        ref variants,
-                                    },
-                            })) => Some(variants),
-                            _ => None,
-                        }
-                    });
+                    match term.value.elements.first() {
+                        Some(&ast::PatternElement::Placeable(
+                            ast::Expression::SelectExpression {
+                                expression: None,
+                                ref variants,
+                            },
+                        )) => Some(variants),
+                        _ => None,
+                    }
+                });
 
                 if let Some(variants) = variants {
                     for variant in variants {
@@ -172,10 +175,7 @@ impl ResolveValue for ast::Expression {
                     return select_default(variants).and_then(|variant| variant.value.to_value(env));
                 }
 
-                message
-                    .as_ref()
-                    .and_then(|message| message.value.as_ref())
-                    .and_then(|pattern| pattern.to_value(env))
+                term.and_then(|term| term.value.to_value(env))
             }
             _ => unimplemented!(),
         }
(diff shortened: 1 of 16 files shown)


src/syntax/ast.rs
```diff
@@ -18,6 +19,14 @@ pub struct Message {
     pub comment: Option<Comment>,
 }
 
+#[derive(Debug, PartialEq)]
+pub struct Term {
+    pub id: Identifier,
+    pub value: Option<Pattern>,
```
stasm: [Terms must have values.](https://github.com/projectfluent/fluent/blob/69ecabcc06cf6f5b3654c439a9defc8575744313/spec/fluent.asdl#L35)

src/syntax/parser.rs
```diff
@@ -72,106 +72,90 @@ where
...
     I: Iterator<Item = char>,
 {
-    let id = get_private_identifier(ps)?;
+    let id = get_entry_identifier(ps)?;
 
-    ps.skip_line_ws();
+    ps.skip_inline_ws();
 
+    // XXX Syntax 0.4 compat
     let pattern = if ps.current_is('=') {
```
stasm: I think we said we wanted to only support the newest Fluent Syntax in `fluent-rs`. Can we use `ps.expect_char('=')` without the `if` here?

tests/resolve_select_expression.rs
```diff
@@ -240,6 +240,7 @@ baz-unknown =
 }
 
 #[test]
+#[ignore]
```
stasm: Why do you need to ignore this test?

Note that [message references cannot be used as selectors](https://github.com/projectfluent/fluent/blob/69ecabcc06cf6f5b3654c439a9defc8575744313/spec/fluent.ebnf#L67-L71) in Syntax 0.5. The `use-foo` message should be updated.

tests/resolve_select_expression.rs
```diff
@@ -273,6 +274,7 @@ use-bar =
...
```
stasm: Why do you need to ignore this test?

stasm:
Github ate half of my review comment: it's great to see the changes to `errors.rs`!

tests/resolve_select_expression.rs
stasm: is this and the other tests ignored because there's currently no way to retrieve terms from `MessageContext`?

src/syntax/parser.rs
zbraniecki: I was hoping that this is the one valve we leave until 1.0 to give us a way out in case our opposition (mine and Pike's) to your decision gains ground?

This is a very small change that I was hoping to keep lax until 1.0 when we'd strengthen the claim.

Is that ok with you?

tests/resolve_select_expression.rs
zbraniecki: yup.

zbraniecki:
do you plan to readd those eaten comments or is it all? :)

zbraniecki:
> I think that for the time being a public MessageContext.get_term should be fine. We should look into how to make it private in the future.

Unfortunately, this won't be enough. The way Rust works, in order for Resolve to be able to handle Expressions it needs to know the shape of the struct, both Message and Term, to get value, attributes and variants (three cases where it needs to access terms).

We have a choice of either:
a) Creating some shared struct that both Term and Message use and `get_message` returns
b) Creating a set of traits that both Term and Message have like `get_attributes`, `get_value`, `get_variants`. This is already partially supported by the `ResolveValue.to_value`.

I don't know of any other way except of manual fallbacking in all three cases.

Do you have a preference?

stasm:
> Unfortunately, this won't be enough. The way Rust works, in order for Resolve to be able to handle Expressions it needs to know the shape of the struct, both Message and Term, to get value, attributes and variants (three cases where it needs to access terms).

I was instead thinking of using `match` guards to tell Messages and Terms apart. Like so:

```diff
diff --git a/src/context.rs b/src/context.rs
index c7a23e5..a663f77 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -35,45 +35,55 @@ use super::resolve::{Env, ResolveValue};
 ///
 /// `MessageContext` stores messages in a single locale, but keeps a locale fallback chain for the
 /// purpose of language negotiation with i18n formatters. For instance, if date and time formatting
 /// are not available in the first locale, `MessageContext` will use its `locales` fallback chain
 /// to negotiate a sensible fallback for date and time formatting.
 #[allow(dead_code)]
 pub struct MessageContext<'ctx> {
     pub locales: &'ctx [&'ctx str],
+    terms: HashMap<String, ast::Term>,
     messages: HashMap<String, ast::Message>,
 }
 
 impl<'ctx> MessageContext<'ctx> {
     pub fn new(locales: &'ctx [&'ctx str]) -> MessageContext {
         MessageContext {
             locales,
+            terms: HashMap::new(),
             messages: HashMap::new(),
         }
     }
 
     pub fn has_message(&self, id: &str) -> bool {
         self.messages.contains_key(id)
     }
 
+    pub fn get_term(&self, id: &str) -> Option<&ast::Term> {
+        self.terms.get(id)
+    }
+
     pub fn get_message(&self, id: &str) -> Option<&ast::Message> {
         self.messages.get(id)
     }
 
     pub fn add_messages(&mut self, source: &str) {
         let res = parse(source).unwrap_or_else(|x| x.0);
 
         for entry in res.body {
             let id = match entry {
+                ast::Entry::Term(ast::Term { ref id, .. }) => id.name.clone(),
                 ast::Entry::Message(ast::Message { ref id, .. }) => id.name.clone(),
                 _ => continue,
             };
 
             match entry {
+                ast::Entry::Term(message) => {
+                    self.terms.insert(id, message);
+                }
                 ast::Entry::Message(message) => {
                     self.messages.insert(id, message);
                 }
                 _ => continue,
             };
         }
     }
 
diff --git a/src/resolve.rs b/src/resolve.rs
index 29336c3..f07e4c2 100644
--- a/src/resolve.rs
+++ b/src/resolve.rs
@@ -76,16 +76,20 @@ impl ResolveValue for ast::VariantName {
 
 impl ResolveValue for ast::Expression {
     fn to_value(&self, env: &Env) -> Option<FluentValue> {
         match *self {
             ast::Expression::StringExpression { ref value } => {
                 Some(FluentValue::from(value.clone()))
             }
             ast::Expression::NumberExpression { ref value } => value.to_value(env),
+            ast::Expression::MessageReference { ref id } if id.name.starts_with('-') => env.ctx
+                .get_term(&id.name)
+                .and_then(|term| term.value.as_ref())
+                .and_then(|pattern| pattern.to_value(env)),
             ast::Expression::MessageReference { ref id } => env.ctx
                 .get_message(&id.name)
                 .and_then(|message| message.value.as_ref())
                 .and_then(|pattern| pattern.to_value(env)),
             ast::Expression::ExternalArgument { ref id } => env.args
                 .and_then(|args| args.get(&id.name.as_ref()))
                 .cloned(),
             ast::Expression::SelectExpression {
diff --git a/tests/resolve_message_reference.rs b/tests/resolve_message_reference.rs
index f378520..cd4fd41 100644
--- a/tests/resolve_message_reference.rs
+++ b/tests/resolve_message_reference.rs
@@ -13,16 +13,31 @@ bar = { foo } Bar
 ",
     );
 
     let value = ctx.get_message("bar").and_then(|msg| ctx.format(msg, None));
     assert_eq!(value, Some("Foo Bar".to_string()));
 }
 
 #[test]
+fn term_reference() {
+    let mut ctx = MessageContext::new(&["x-testing"]);
+
+    ctx.add_messages(
+        "
+-foo = Foo
+bar = { -foo } Bar
+",
+    );
+
+    let value = ctx.get_message("bar").and_then(|msg| ctx.format(msg, None));
+    assert_eq!(value, Some("Foo Bar".to_string()));
+}
+
+#[test]
 fn message_reference_nested() {
     let mut ctx = MessageContext::new(&["x-testing"]);
 
     ctx.add_messages(
         "
 foo = Foo
 bar = { foo } Bar
 baz = { bar } Baz
```

stasm:
> a) Creating some shared struct that both Term and Message use and get_message returns

I'd like to avoid the situation in which `get_message` can return a Term.

> b) Creating a set of traits that both Term and Message have like get_attributes, get_value, get_variants. This is already partially supported by the ResolveValue.to_value.

This sounds like a good refactoring that we could follow up with independently of this PR!

stasm:
> do you plan to readd those eaten comments or is it all? :)

Sorry, I didn't express myself well enough. The eaten comments were about the `errors.rs` file. Rather than reproduce them, I just went for the happy _it's great to see the changes to errors.rs!_ in my comment above.

src/syntax/parser.rs
stasm: No, I think we should stick to the Syntax spec as close as possible. As you say, it's a small change—it's easy to bring it back if we need to. I don't see any value in keeping the double parsing behavior around in `fluent-rs` at this point.

zbraniecki:
Enforced the `=` and added the 0.6.2 parser patch. @stasm - ready for your review again :)

src/resolve.rs
```diff
@@ -127,10 +130,16 @@ impl ResolveValue for ast::Expression {
...
-                    .as_ref()
-                    .and_then(|message| message.attributes.as_ref());
+                let attributes = if id.name.starts_with('-') {
+                    env.ctx
+                        .get_term(&id.name)
+                        .and_then(|message| message.attributes.as_ref())
+                } else {
+                    env.ctx
+                        .get_message(&id.name)
+                        .as_ref()
```
stasm: I suspect that in general this file has a few `as_ref`s too many :) I think we can follow-up with a clean up later on, but in this particular case I'd like to see the same code in both branches of this `if-else` please.

PR #46, opened 2018-02-21
clippy: Don't use return where it isn't needed.

waywardmonkeys:


zbraniecki:
Thank you for the contribution!

Issue #25, opened 2017-10-11
Support custom functions in MessageContext


zbraniecki:
We want to support passing arbitrary generic functions as arguments to MessageContext constructor to allow developers to provide builtins into localization contexts.

JohnDoneth:
Close due to #50?

zbraniecki:
Yep! :) Thanks again!

stasm:
Nice, that was a big one. It feels great to be able to cross it off the list!

PR #50, opened 2018-05-02
Support custom functions in MessageContext #25

JohnDoneth:
Request for Review.

Some questions arose while implementing this feature.

- Should the custom functions be able to return `None` or always return a `FluentValue`?
- Should the custom functions be `FnMut` or be stateless?
- Can this same code be implemented using non-boxed closures? The `Box` is a bit of an eyesore but is needed due to a `Sized` trait requirement for storing the functions?

Otherwise, the code works as intended according to the [functions portion of the Project Fluent spec](http://projectfluent.org/fluent/guide/functions.html). Please let me know if I missed anything or how to improve this pull request before merging. 👍

### DIFF
--- a/examples/functions.rs
+++ b/examples/functions.rs
@@ -1,7 +1,7 @@
 extern crate fluent;
 
-use fluent::MessageContext;
 use fluent::types::FluentValue;
+use fluent::MessageContext;
 
 fn main() {
     let mut ctx = MessageContext::new(&["en-US"]);
@@ -38,18 +38,21 @@ fn main() {
     ctx.add_messages("meaning-of-life = { MEANING_OF_LIFE(42) }");
     ctx.add_messages("all-your-base = { BASE_OWNERSHIP(hello, ownership: \"us\") }");
 
-    let value = ctx.get_message("hello-world")
+    let value = ctx
+        .get_message("hello-world")
         .and_then(|message| ctx.format(message, None));
     assert_eq!(value, Some("Hey there! I'm a function!".to_string()));
 
-    let value = ctx.get_message("meaning-of-life")
+    let value = ctx
+        .get_message("meaning-of-life")
         .and_then(|message| ctx.format(message, None));
     assert_eq!(
         value,
         Some("The answer to life, the universe, and everything".to_string())
     );
 
-    let value = ctx.get_message("all-your-base")
+    let value = ctx
+        .get_message("all-your-base")
         .and_then(|message| ctx.format(message, None));
     assert_eq!(value, Some("All your base belong to us".to_string()));
 }
--- a/examples/functions.rs
+++ b/examples/functions.rs
@@ -7,41 +7,32 @@ fn main() {
     let mut ctx = MessageContext::new(&["en-US"]);
 
     // Test for a simple function that returns a string
-    ctx.add_function(
-        "HELLO",
-        Box::new(|_args, _named_args| {
-            return Some("I'm a function!".into());
-        }),
-    );
+    ctx.add_function("HELLO", |_args, _named_args| {
+        return Some("I'm a function!".into());
+    });
 
-    // Test for a function that accepts unnamed arguments
-    ctx.add_function(
-        "MEANING_OF_LIFE",
-        Box::new(|args, _named_args| {
-            if args.len() > 0 {
-                if args[0] == FluentValue::Number(42.0) {
-                    return Some("The answer to life, the universe, and everything".into());
-                }
+    // Test for a function that accepts unnamed positional arguments
+    ctx.add_function("MEANING_OF_LIFE", |args, _named_args| {
+        if let Some(arg0) = args.get(0) {
+            if *arg0 == Some(FluentValue::Number(42.0)) {
+                return Some("The answer to life, the universe, and everything".into());
             }
+        }
 
-            None
-        }),
-    );
+        None
+    });
 
     // Test for a function that accepts named arguments
-    ctx.add_function(
-        "BASE_OWNERSHIP",
-        Box::new(|_args, named_args| {
-            let ownership = named_args.get("ownership").unwrap();
+    ctx.add_function("BASE_OWNERSHIP", |_args, named_args| {
+        let ownership = named_args.get("ownership").unwrap();
 
-            return match ownership {
-                &FluentValue::String(ref string) => {
-                    Some(format!("All your base belong to {}", string).into())
-                }
-                _ => None,
-            };
-        }),
-    );
+        return match ownership {
+            &FluentValue::String(ref string) => {
+                Some(format!("All your base belong to {}", string).into())
+            }
+            _ => None,
+        };
+    });
 
     ctx.add_messages("hello-world = Hey there! { HELLO() }");
     ctx.add_messages("meaning-of-life = { MEANING_OF_LIFE(42) }");
--- a/examples/functions.rs
+++ b/examples/functions.rs
@@ -1,0 +1,70 @@
+extern crate fluent;
+
+use fluent::MessageContext;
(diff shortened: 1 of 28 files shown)


JohnDoneth:
Working on fixing the Travis build for stable rust at the moment.

zbraniecki:
This looks awesome! I'll take on the review this week!

JohnDoneth:
Additional things to improve in the future to keep this pull request small:
- Implement the NUMBER function with implicit use
- Implement the DATETIME function, possibly via the [Rust chrono library](https://github.com/chronotope/chrono)? Which should support all of the time related features, including time-zones, required to fulfill all of the different parameters via the builder pattern.

stasm:
>> Should the custom functions be able to return None or always return a FluentValue?

> I believe so. @stasm - do you agree?

_(Which one are you asking about?)_ 😉

I like the solution from the PR, i.e. custom functions return `Option<FluentValue>`. `ResolveValue` already knows how to handle `Options` in the implementations for `Pattern` (when a `Placeable` might be `None`) and `SelectExpression` (when the `selector` might be `None`):

https://github.com/projectfluent/fluent-rs/blob/d020470836e8307d9fa22b42b2096209e0a4ed7f/src/resolve.rs#L49-L60

https://github.com/projectfluent/fluent-rs/blob/d020470836e8307d9fa22b42b2096209e0a4ed7f/src/resolve.rs#L103-L130

stasm:
I should also mention that the mid-term plan is to switch to using `Result` rather than `Option`. We'll likely want to do the same for custom functions, which would be considered a breaking API change.

src/resolve.rs
```diff
@@ -177,6 +177,44 @@ impl ResolveValue for ast::Expression {
...
+            } => {
+
+                let mut resolved_unnamed_args = Vec::new();
+                let mut resolved_named_args = HashMap::new();
+
+                for arg in args {
+                    match arg {
+                        &ast::Argument::Expression(ref expression) => {
+                            if let Some(resolved_expression) = expression.to_value(env) {
+                                resolved_unnamed_args.push(resolved_expression);
```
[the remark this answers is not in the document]
JohnDoneth: Ah yes that makes sense. Thank you for catching that!

src/context.rs
```diff
@@ -65,6 +70,27 @@ impl<'ctx> MessageContext<'ctx> {
         self.terms.get(id)
     }
 
+    pub fn add_function(
+        &mut self,
+        id: &str,
+        func: Box<Fn(&[FluentValue], &HashMap<String, FluentValue>) -> Option<FluentValue>>,
+    ) {
+        self.functions.insert(id.to_string(), func);
```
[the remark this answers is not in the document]
JohnDoneth: Awesome! This is what I wanted to do but I had forgotten you can use Fn like a trait requirement because well... it is a trait.

src/context.rs
```diff
@@ -65,6 +70,23 @@ impl<'ctx> MessageContext<'ctx> {
         self.terms.get(id)
     }
 
+    pub fn add_function<F>(&mut self, id: &str, func: F)
+    where
+        F: 'static
```
stasm: is there a reason for using `'static` rather than `'ctx`?

stasm:
@zbraniecki I think this is mostly waiting for your review. Want to take a look?

src/context.rs
JohnDoneth: I was unable to get `'ctx` to compile

stasm: If it helps, the following changes worked for me:

```diff
diff --git a/src/context.rs b/src/context.rs
index 32ab07a..1caaf69 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -39,17 +39,19 @@ use super::resolve::{Env, ResolveValue};
 /// to negotiate a sensible fallback for date and time formatting.
 #[allow(dead_code)]
 pub struct MessageContext<'ctx> {
     pub locales: &'ctx [&'ctx str],
     messages: HashMap<String, ast::Message>,
     terms: HashMap<String, ast::Term>,
     functions: HashMap<
         String,
-        Box<Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>>,
+        Box<
+            'ctx + Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>,
+        >,
     >,
 }
 
 impl<'ctx> MessageContext<'ctx> {
     pub fn new(locales: &'ctx [&'ctx str]) -> MessageContext {
         MessageContext {
             locales,
             messages: HashMap::new(),
@@ -67,27 +69,28 @@ impl<'ctx> MessageContext<'ctx> {
     }
 
     pub fn get_term(&self, id: &str) -> Option<&ast::Term> {
         self.terms.get(id)
     }
 
     pub fn add_function<F>(&mut self, id: &str, func: F)
     where
-        F: 'static
-            + Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>,
+        F: 'ctx + Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>,
     {
         self.functions.insert(id.to_string(), Box::new(func));
     }
 
     pub fn get_function(
         &self,
         id: &str,
     ) -> Option<
-        &Box<Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>>,
+        &Box<
+            'ctx + Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>,
+        >,
     > {
         self.functions.get(id)
     }
 
     pub fn add_messages(&mut self, source: &str) {
         let res = parse(source).unwrap_or_else(|x| x.0);
 
         for entry in res.body {

```

examples/functions.rs
```diff
@@ -0,0 +1,58 @@
...
+    let mut ctx = MessageContext::new(&["en-US"]);
+
+    // Test for a simple function that returns a string
+    ctx.add_function("HELLO", |_args, _named_args| {
+        return Some("I'm a function!".into());
+    });
+
+    // Test for a function that accepts unnamed positional arguments
+    ctx.add_function("MEANING_OF_LIFE", |args, _named_args| {
+        if let Some(arg0) = args.get(0) {
```
zbraniecki: You can use a `match` here instead of a nested ifs.

src/context.rs
```diff
@@ -65,6 +72,24 @@ impl<'ctx> MessageContext<'ctx> {
         self.terms.get(id)
     }
 
+    pub fn add_function<F>(&mut self, id: &str, func: F)
+    where
+        F: 'ctx + Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>,
+    {
+        self.functions.insert(id.to_string(), Box::new(func));
+    }
```
zbraniecki: In JS we pass functions into the constructor of the `MessageContext`. Is there any reason you moved it to a separate function? I'd prefer to stay consistent with JS when possible.

stasm: Actually, I'd argue that `add_function` is a good design for Rust. It's consistent with how you populate `HashMaps` via `insert()`. And it reduces the boilerplate. The alternative would be to pass a `HashMap` of functions into the constructor. In order to build the `HashMap` you'd need to `insert()` every function separately anyways. `add_function` abstracts that away.

zbraniecki: oh, ok! Convinced!

zbraniecki:
In the light of @stasm comment, I'm going to merge this PR now!

src/context.rs
JohnDoneth: I agree with @stasm. I tried making the changes that you suggested yesterday @zbraniecki but I encountered the same issues that @stasm brought up.

PR #55, opened 2018-07-21
Resolve Cyclic Reference Issue

### LINKED ISSUE #20: ResolveValue::to_value should return a Result
Right now it returns an `Option`. By returning a `Result` instead we could capture error messages and fallback values.

### LINKED ISSUE #21: Protect against cyclic references
See the skipped test added in 165a75c0480ae2642ec9f6ca6fcce607f3ef46ef.

mmstick:
- `MessageContext` now uses a single map for messages, terms, functions
- `ResolveValue::to_value` now returns a Result, required to handle cyclic resolves
- Cyclic resolves are no longer an issue, due to tracking previously-tracked identifiers
- An extra test was added for resolving a message multiple times within the same expression

Closes #20 
Closes #21

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -5,14 +5,15 @@
 //! together.
 
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::hash_map::{Entry as HashEntry, HashMap};
 
+use super::errors::FluentError;
 use super::resolve::{Env, ResolveValue};
 use super::syntax::ast;
 use super::syntax::parse;
 use super::types::FluentValue;
 
-enum Variant<'ctx> {
+enum Entry<'ctx> {
     Message(ast::Message),
     Term(ast::Term),
     Function(
@@ -51,7 +52,7 @@ enum Variant<'ctx> {
 #[allow(dead_code)]
 pub struct MessageContext<'ctx> {
     pub locales: &'ctx [&'ctx str],
-    map: HashMap<String, Variant<'ctx>>,
+    map: HashMap<String, Entry<'ctx>>,
 }
 
 impl<'ctx> MessageContext<'ctx> {
@@ -64,7 +65,7 @@ impl<'ctx> MessageContext<'ctx> {
 
     pub fn has_message(&self, id: &str) -> bool {
         self.map.get(id).map_or(false, |id| {
-            if let Variant::Message(_) = id {
+            if let Entry::Message(_) = id {
                 true
             } else {
                 false
@@ -74,7 +75,7 @@ impl<'ctx> MessageContext<'ctx> {
 
     pub fn get_message(&self, id: &str) -> Option<&ast::Message> {
         self.map.get(id).and_then(|id| {
-            if let Variant::Message(ref msg) = id {
+            if let Entry::Message(ref msg) = id {
                 Some(msg)
             } else {
                 None
@@ -84,7 +85,7 @@ impl<'ctx> MessageContext<'ctx> {
 
     pub fn get_term(&self, id: &str) -> Option<&ast::Term> {
         self.map.get(id).and_then(|id| {
-            if let Variant::Term(ref term) = id {
+            if let Entry::Term(ref term) = id {
                 Some(term)
             } else {
                 None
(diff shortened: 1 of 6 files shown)


zbraniecki:
Thanks Michael! Happy to see the cyclic reference issue handled!

@stasm - what's your take on using a single map for messages/terms/functions? It feels weird but it may be just that I didn't think about it enough.

stasm:
I think I like it. It makes the code simpler in some parts, at the cost of requiring runtime checks in `has_message` and `get_message`. Which is probably fine. We might want to look at https://github.com/projectfluent/fluent.js/issues/208 and see how it will look like in `fluent-rs` using the single map approach.

mmstick:
I can remove the single map and place that in another PR. It's not required to solve this issue. A single map is how we handle variables within Ion, which may be functions, aliases, strings, arrays, hash maps, btree maps, etc. You usually don't want variables with overlapping names.

src/context.rs
```diff
@@ -4,13 +4,24 @@
...
 
+use std::cell::RefCell;
 use std::collections::HashMap;
 
 use super::resolve::{Env, ResolveValue};
 use super::syntax::ast;
 use super::syntax::parse;
 use super::types::FluentValue;
 
+enum Variant<'ctx> {
```
zbraniecki: Can you call it `Entry` please? Variants are used in FTL as a public concept. Entries are in AST but are just a bag of message/comment/junk, so I'm ok having another Entry in Context that is a bad of message/term/function.

src/context.rs
```diff
@@ -40,43 +51,53 @@ use super::types::FluentValue;
...
+        })
     }
 
     pub fn add_function<F>(&mut self, id: &str, func: F)
     where
         F: 'ctx + Fn(&[Option<FluentValue>], &HashMap<String, FluentValue>) -> Option<FluentValue>,
     {
-        self.functions.insert(id.to_string(), Box::new(func));
+        self.map
+            .insert(id.to_string(), Variant::Function(Box::new(func)));
```
zbraniecki: Can you guard here against already existing entry and throw if encountered? (it probably should make `add_function` return a result)

src/context.rs
```diff
@@ -102,10 +129,10 @@ impl<'ctx> MessageContext<'ctx> {
 
             match entry {
                 ast::Entry::Message(message) => {
-                    self.messages.insert(id, message);
+                    self.map.insert(id, Variant::Message(message));
```
zbraniecki: Same here. Guard against overriding by throwing - https://github.com/projectfluent/fluent.js/blob/master/fluent/src/context.js#L153

src/resolve.rs
```diff
@@ -162,19 +232,22 @@ impl ResolveValue for ast::Expression {
...
+                    .to_value(env)
             }
             ast::Expression::CallExpression {
                 ref callee,
                 ref args,
             } => {
-                let mut resolved_unnamed_args = Vec::new();
-                let mut resolved_named_args = HashMap::new();
+                let resolved_unnamed_args = &mut Vec::new();
+                let resolved_named_args = &mut HashMap::new();
```
zbraniecki: Is that now a preferred convention?

tests/resolve_message_reference.rs
```diff
@@ -73,3 +72,18 @@ bar = { foo } Bar
...
+    ctx.add_messages(
+        "
+foo = Foo
+bar = { foo } Bar { foo }
+",
+    );
+
+    let value = ctx.get_message("bar").and_then(|msg| ctx.format(msg, None));
+    assert_eq!(value, Some("Foo Bar Foo".to_string()));
+}
```
zbraniecki: Can you add a test that shows that we're breaking out of a cyclic reference?

```
foo = { bar }
bar = { foo }
```

mmstick: Sure thing

src/resolve.rs
mmstick: It's less a convention and more of a requirement in some situations, when you want to ensure that the borrow checker understands that you only want to pass a reference to a closure, rather than having the closure take ownership of the value you want to reference within.

It can be more idiomatic in some situations though, when you know that you're going to just borrow the thing repeatedly.

zbraniecki:
Thank you! Last question - how opinionated your are that we should include the failure/failure_derive crates? I'm concerned about dependency creep and since I never used this one I'm not sure, after skimming through its docs, if they add that much value.

Would it be hard to avoid using them?

mmstick:
It's a pretty popular crate in the Rust community. The author & maintainer of the crate, "withoutboats", is a [member of the Rust language team](https://www.rust-lang.org/en-US/team.html). I can remove it and implement the `Display` trait by hand, though.

We use it in all of our projects at System76, as well as in Redox OS. It makes constructing errors a lot easier, as there's no need to manually implement the `Display` trait separately. It especially improves the QoL of a project that defines many error types, or has error types with many possible elaborate variants. You define how the error message is to be constructed for each variant in the enum in-line within the enum definition.

Examples from some of our projects:

- [distinst](https://github.com/pop-os/distinst/blob/master_bionic/src/disk/error.rs)
- [distinst-cli](https://github.com/pop-os/distinst/blob/4563f7681942d0bbe42525171d554d205e1202c5/cli/src/errors.rs)
- [debrepobuilder](https://github.com/pop-os/debrepbuild/search?utf8=%E2%9C%93&q=%23%5Bderive%28Debug%2C+Fail%29%5D&type=)

mmstick:
Apparently, it's located in the Rust Language Nursery now: https://github.com/rust-lang-nursery/failure

zbraniecki:
Convinced, thank you!

PR #56, opened 2018-07-31
Plural rule handling from intl_pluralrules

unclenachoduh:
Remove mock implementation for plural rules and integrate intl_pluralrules crate.


fluent/src/types.rs
```diff
@@ -40,10 +45,32 @@ impl FluentValue {
...
+                let supported = negotiate_languages(
+                    ctx.locales,
+                    IntlPluralRules::get_locales(PluralRuleType::CARDINAL),
+                    Some("en"),
+                    &NegotiationStrategy::Lookup,
+                );
+
+                let locale = supported[0].to_owned();
+
+                let pr = IntlPluralRules::create(&locale, PluralRuleType::CARDINAL).unwrap();
```
zbraniecki: Turning a string slice into an owner String, just to pass a reference (slice) to it in the next line doesn't make much sense ;)
Next time:

```
let pr = IntlPluralRules::create(supported[0], ...);
```

fluent/Cargo.toml
```diff
@@ -22,3 +22,4 @@ fluent-locale = "0.3.1"
 fluent-syntax = "0.1.0"
 failure = "0.1.1"
 failure_derive = "0.1.1"
+intl_pluralrules = "0.8.1"
```
bjorn3: Missing trailing newline

PR #63, opened 2018-08-09
MessageContext.format now takes an id

### LINKED ISSUE #60: Remove MessageContext.get_message
This is similar to https://github.com/projectfluent/fluent.js/issues/208.

I'd like to change `MessageContext.format()` to accept an identifier of a message and format it directly to a string. It would avoid the intermediate step of having to first `get_message` the message from the context, which is a bit clunky and is an artifact of how the JavaScript bindings retrieve translated attributes.

stasm:
- Make `get_message` private.
- Remove `get_term` and `get_function`.
- `format` must now be called with the string id of the desired message:

        ctx.format("foo", None);

    It also accepts dot-separated paths identifying attributes:

        ctx.format("foo.attrname", None);

Fix #60.

### DIFF
--- a/fluent/examples/external_arguments.rs
+++ b/fluent/examples/external_arguments.rs
@@ -1,12 +1,11 @@
 extern crate fluent;
 
-use fluent::context::MessageContext;
+use fluent::MessageContext;
 use fluent::types::FluentValue;
 use std::collections::HashMap;
 
 fn main() {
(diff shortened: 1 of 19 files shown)


stasm:
I didn't want to make too many changes in this PR, so I kept the return type of `format` as `Option<String>`. We might want to change it to `Result` in the future.

PR #68, opened 2018-09-23
Specific lifetime on initialiser

markpritchard:
When not explicitly defined, Rust will assign the same lifetime to all
input parameters and return values.

The FluentBundle::new constructor does not define an explicit lifetime
for the 'locale' input parameter, so it is assigned the same lifetime as
the bundle itself. This causes the borrow checker to reject code that
initialises bundles populated from dynamically loaded resources and
stored in separate containers.

This adds an explicit lifetime to the locales parameter to decouple it
from the return value and a test case that fails to compile without the
change.

### DIFF
--- a/fluent/src/bundle.rs
+++ b/fluent/src/bundle.rs
@@ -56,7 +56,8 @@ pub struct FluentBundle<'bundle> {
 }
 
 impl<'bundle> FluentBundle<'bundle> {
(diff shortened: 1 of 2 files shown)


kirhgoff:
This example reproduces the issue: https://github.com/kirhgoff/fluent-rs-lifetime-issue

Issue #16, opened 2017-08-17
Consider rewriting the parser to make it copy-free


zbraniecki:
At the moment the parser is copying some parts of the slice while creating the AST.

Most Rust parsers I can find in the wild rather rely on building AST that stores slices and aim for zero-copy mode.
Last example: https://github.com/viperscape/lichen/blob/master/src/parse.rs

I think we could do this here and it would probably benefit performance and memory. The cost is that if you want to modify the AST you'll have to copy it, but that seems to be a reasonable cost.
Most use cases only need to read the AST from what I know.

Cyres:
Basically every `String` could be replaced with `&'ctx str` as far as i inspected the source code.

zbraniecki:
Fixed by https://github.com/projectfluent/fluent-rs/pull/76

PR #76, opened 2018-12-21
Refactor of the code, 0.8 compatibility, zero-copy AST/parsing, Rust 2018 goodies

zbraniecki:
This is a major rewrite, to the point where I'm not even sure how to exactly plan reviewing it and landing, but I think it is getting close to being done.

There's one open question about `FluentBundle::add_messages` which I posted in https://github.com/projectfluent/fluent.js/issues/328

This, unfortunately, break a lot of examples/tests in fluent-bundle as before they freely passed `&str` and constructed `FluentResource` on fly.
We'll need to answer that question, and then clean up the code, make sure that such zero-copy abstraction works well for Amethyst, and begin reviewing it to land.

I believe that the proposed structure and memory allocation model should bring us close to the model we'll use for fluent-rs 1.0.


zbraniecki:
My initial thinking is to split this PR into 4 PR's each updating a separate crate and get them reviewed separately, releasing them in order `fluent-syntax` -> `fluent-cli` -> `fluent-bundle` -> `fluent`.

zbraniecki:
For `fluent`'s `ResourceManager` I'll want to migrate it to use Manish's elsa - https://github.com/Manishearth/elsa

zbraniecki:
This also resolves https://github.com/projectfluent/fluent-rs/issues/16

stasm:
> My initial thinking is to split this PR into 4 PR's each updating a separate crate and get them reviewed separately, releasing them in order fluent-syntax -> fluent-cli -> fluent-bundle -> fluent.

What is the benefit of going for 4 separate packages? In my experience, too much granularity makes publishing new versions tedious and it also means that there are more permutations of packages installed in the environment at the same time, which makes it easy to run into errors due to mismatched dependencies.

I can see the benefit of publishing `fluent-cli` as a separate package, since that's a _bin_-type crate, rather than _lib_. I would recommend putting the other three in a single `fluent` crate.

zbraniecki:
> What is the benefit of going for 4 separate packages?

Each package has different set of requirements and "adds up". The `fluent-syntax`, much like in case of JS and python will be useful for tooling environments which don't need anything else beyond `parse/ast/serialize` cycle.
`fluent-bundle` adds the resolver but remains independent of any I/O making it the right level for environments such as gaming systems or Gecko which provide their own resource management systems.
`fluent` is the `complete` package which provides its own, minimal, ResourceManager minimizing the amount of custom code one has to write to make their project localizable.
`fluent-cli` (or maybe `fluent-cli-tools`?) is a set of binaries to work with ftl/fluent which may eventually become an umbrella for a collection of separately released tools (`fluent-cli-parse`, `fluent-cli-validator` etc.).

Does it make sense?

> In my experience, too much granularity makes publishing new versions tedious and it also means that there are more permutations of packages installed in the environment at the same time, which makes it easy to run into errors due to mismatched dependencies.

I think Rust cargo is really good with that and the ecosystem actively encourages dividing into many small crates which allows other to better pick which parts they need (see the family of `serde` crates) and allows for parallel compilation of the crates.

> I would recommend putting the other three in a single fluent crate.

It would mean that any tool would have to compile the Bundle/Resolver/ResourceManager parts of Fluent even if all they need it FTL parser. I think that this use case is distinct enough to warrant splitting it out (much like librustc uses `syntax`, `syntax_pos` and even `serialize` as separate crates - https://github.com/rust-lang/rust/blob/master/src/librustc/Cargo.toml#L28).

As to if `fluent-bundle` should be separate crate, I'm less strongly opinionated.

zbraniecki:
Hooray! This branch now compiles and all tests pass! The performance is good, but can be better with time.

Initially, with no optimization work put into the parser, I'm seeing:

master:
```
test bench_menubar_parse ... bench:     407,840 ns/iter (+/- 32,920)
test bench_simple_parse  ... bench:      45,287 ns/iter (+/- 4,698)
```

zero-copy-resolver:
```
test bench_menubar_parse ... bench:     155,191 ns/iter (+/- 31,100)
test bench_simple_parse  ... bench:      22,147 ns/iter (+/- 15,138)
```

the resolver benchmarks are preserved.

I'd like to land this update soon now, to bring the rust implementation to Fluent 0.8 syntax and move to work on documentation, errors, ergonomics and higher level bits (like ResourceManager).

zbraniecki:
I asked @ManishEarth to review the examples and crate split and he said that it seems like a good natural split of meta-crates.

I'd like to preserve the split between `fluent` and `fluent-bundle` and eventually consider moving the `ResourceManager` to its own crate like `fluent-resource-manager` and referencing it from the catch-all `fluent` crate.

zbraniecki:
@Xaeroxe - can you verify that the `zero-copy-resolver` branch works with Amethyst model? In particular I'd like to make sure that:

* The zero-copy parser into `fluent-bundle::resource::FluentResource` which uses `rental` to allocate its `String` works well.
* The API changes I'm making in `fluent-bundle` which allow libs like Amethyst to plug their own resource manager will work for `amethyst-locale`.
* The POC of the `fluent::resource_manager::ResourceManager` is something that you'll either want to use, or replace with your own resource manager.

Thank you!

fluent-bundle/src/resource.rs
```diff
@@ -0,0 +1,40 @@
...
+            ast: ast::Resource<'string>,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct FluentResource(rentals::FluentResource);
+
+impl FluentResource {
+    pub fn try_new(source: String) -> Result<Self, (Self, Vec<ParserError>)> {
```
JohnHeitmann: Can/should this take Into<String> for better caller ergonomics?

fluent-bundle/src/bundle.rs
```diff
@@ -167,20 +152,18 @@ impl<'bundle> FluentBundle<'bundle> {
...
+                    match attribute.to_value(&env) {
+                        Ok(val) => {
+                            return Some((val.format(self), errors));
+                        }
+                        Err(err) => {
+                            errors.push(FluentError::ResolverError(err));
+                            // XXX: In the future we'll want to get the partial
+                            // value out of resolver and return it here.
+                            // We also expect to get a Vec or errors out of resolver.
+                            return Some((path.to_string(), errors));
```
JohnHeitmann: Can `format_message` be removed as part of this change? It looks like `format` does everything a caller might want. If not I'll add it to my doc pr.

fluent-bundle/src/bundle.rs
```diff
@@ -57,10 +57,7 @@ pub struct FluentBundle<'bundle> {
 
 impl<'bundle> FluentBundle<'bundle> {
     pub fn new<'a, S: ToString>(locales: &'a [S]) -> FluentBundle<'bundle> {
-        let locales = locales
-            .into_iter()
-            .map(|s| s.to_string())
-            .collect::<Vec<_>>();
+        let locales = locales.iter().map(|s| s.to_string()).collect::<Vec<_>>();
```
JohnHeitmann: Should FluentBundle.{locales,entries,plural_rules} be made private as part of this change?

fluent-bundle/src/resource.rs
zbraniecki: We considered that with @manishearth and he recommended not doing that in order to avoid hiding the allocation.
While ergonomically that would be nice, it would also make it harder for the reader to understand that we will allocate the String and store it inside of the FluentResource.

In the future, when static `&str` will become implicitly convertible to String [0] it should be easier.

[0] https://github.com/rust-lang/rust/pull/46993

fluent-bundle/src/bundle.rs
zbraniecki: It doesn't do everything. `format` allows you to get the message's value or a single attribute, but it doesn't allow you to retrieve value+attributes together.

`format_message` will be useful for higher level crates that will apply a localization message onto a widget/element.

fluent-bundle/src/resource.rs
Manishearth: Note: a thing that could be done here is to make the return value tuple into a proper public struct so it's easier to document

fluent-bundle/src/bundle.rs
zbraniecki: Seems like `entries` and `plural_rules` need to stay public to be usable from outside of the struct. We could limit `locales`, but I'm wondering if we should invest in some API design.

fluent-syntax/src/ast.rs
```diff
@@ -1,129 +1,134 @@
...
+    pub name: &'ast str,
 }
 
 #[derive(Debug, PartialEq)]
-pub enum Argument {
-    Expression(Expression),
-    NamedArgument { name: Identifier, val: ArgValue },
+pub struct Variant<'ast> {
+    pub key: VariantKey<'ast>,
+    pub value: Value<'ast>,
```
Pike: This is a pattern in the ebnf?

fluent-syntax/src/ast.rs
```diff
@@ -1,129 +1,134 @@
...
+    },
+    MessageReference {
+        id: Identifier<'ast>,
+    },
+    TermReference {
+        id: Identifier<'ast>,
+    },
+    FunctionReference {
+        id: Identifier<'ast>,
+    },
```
Pike: The `FunctionReference` isn't in the ebnf?

fluent-syntax/src/ast.rs
zbraniecki: Good catch! Thank you!

fluent-syntax/src/ast.rs
zbraniecki: The issue is in :

```ebnf
CallExpression ::= CalleeExpression blank? "(" blank? argument_list blank? ")"
CalleeExpression    ::= AttributeExpression
                      | FunctionReference
                      | TermReference
```

The `callee` field of `CallExpression` must be *something* and that something is at the moment made of two `InlineExpression` variants (`AttributeExpression` and `TermReference`) and a standalone `FunctionReference`.
What's more `FunctionReference` is not used anywhere else.

So, my options are:

*) Create a standalone node called `FunctionReference` and create a `CalleeExpression` enum that can take either `InlineExpression` or `FunctionReference` and continue guarding against invalid variants of `InlineExpression` being used in the `callee` field in the parser rather than ast.
*) Move both `AttributeExpression` and `TermReference` to be standalone Nodes that can be also used as variants of InlineExpression and then create `CalleeExpression` to reference them.
*) Do what I did here

The second option would look sth like this:

```rust
struct AttributeExpression {...}
struct TermReference {...}
struct FunctionReference {...}

enum InlineExpression {
  StringLiteral{ raw: &str },
  NumberLiteral{ value: &str },
  VariableReference{ id: Identifier },
  CallExpression {...},
  AttributeExpression(AttributeExpression),
  VariantExpression { ... },
  MessageReference { ... },
  TermReference(TermReference),
  Placeable { ... }
}

enum CalleeExpression {
  AttributeExpression(AttributeExpression),
  TermReference(TermReference),
  FunctionReference(FunctionReference), 
}
```

I think it's not worth the trouble. I'll comment in the AST.

zbraniecki:
I was able to place some small optimizations around the parser, and the current numbers are:

master:
```
test bench_menubar_parse ... bench:     407,840 ns/iter (+/- 32,920)
test bench_simple_parse  ... bench:      45,287 ns/iter (+/- 4,698)
```

branch:
```
test bench_menubar_parse ... bench:     127,473 ns/iter (+/- 5,210)
test bench_simple_parse  ... bench:      19,919 ns/iter (+/- 3,886)
```

What's cool is that all of that was primarily me removing code that didn't check if the pointer is in range of the string preparing for the fuzzer. The perf win is a side effect. :)

stasm:
> Each package has different set of requirements and "adds up". The fluent-syntax, much like in case of JS and python will be useful for tooling environments which don't need anything else beyond parse/ast/serialize cycle.
fluent-bundle adds the resolver but remains independent of any I/O making it the right level for environments such as gaming systems or Gecko which provide their own resource management systems.
fluent is the complete package which provides its own, minimal, ResourceManager minimizing the amount of custom code one has to write to make their project localizable.
fluent-cli (or maybe fluent-cli-tools?) is a set of binaries to work with ftl/fluent which may eventually become an umbrella for a collection of separately released tools (fluent-cli-parse, fluent-cli-validator etc.).

> Does it make sense?

Yes, it makes sense, thanks. My question was driven by pragmatism and the following question: do we have use-cases for so much granularity? Sometimes it's better to design hierarchies based on the observed use-cases rather than on how things derive from each other logically.

However, since you already did all the work to separate them, it's probably best to keep them that way.

> I think Rust cargo is really good with that and the ecosystem actively encourages dividing into many small crates which allows other to better pick which parts they need (see the family of serde crates) and allows for parallel compilation of the crates.

This isn't as much about cargo as it is about the rigor to always update the readmes, the changelogs, the dependencies of higher-level packages etc. Having more packages to juggle also means that it's easier to create permutations of versions which we haven't thought about.

fluent-bundle/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+extern crate rental;
+#[macro_use]
+extern crate failure_derive;
+
+pub mod bundle;
+pub mod entry;
+pub mod errors;
+pub mod resolve;
+pub mod resource;
+pub mod types;
```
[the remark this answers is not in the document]
zbraniecki: I wanted to wait with such convenience parts until the code stabilizes a little bit. In my mind it's easier to add it than to remove it.
Do you feel like those three structs are certain to be the ones we want to expose on the main crate?

fluent-syntax/src/parser/ftlstream.rs
```diff
@@ -1,394 +1,211 @@
...
         }
     }
 
-    fn take_digit(&mut self) -> Option<char> {
-        let closure = |x| match x {
-            '0'...'9' => true,
-            _ => false,
-        };
+    pub fn get_slice(&self, start: usize, end: usize) -> &'p str {
+        unsafe { str::from_utf8_unchecked(&self.source[start..end]) }
```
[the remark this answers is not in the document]
zbraniecki: The consequence is that I'm using our collective brain power, instead of compiler's brain power, to ensure that I'm not slicing outside of the boundary.
The boundary checks are relatively expensive and since we're a full byte parser I believe we always attempt to slice within the range.

So, it's not about whether the input is valid UTF8, it's only about whether the parser always selects the right start/end combo that's within the source slice.

I'm going to use a fuzzer to further verify that claim, but I'd prefer not to overblow the lib with options that don't really mean too much to the user. Currently if you pick wrong start/end, the parser crashes, but if we instead use `from_utf8` we'll likely use it with `expect` which will panic as well :)

If while working on 1.0 we'll discover that we have a lot of places in the parser that pick start/end wrong we can try to mitigate it here, but I'd honestly prefer to fix them in the parser.

Please, not that unsafe code is not "bad" per design. It makes total sense to use it in many cases, and people like Henri are actively asking for improvements in how we calculate boundary check requirements to minimize when we need them: https://hsivonen.fi/rust2019/ .

Our problem is that we have code like:

```rust
let source = "id = Foo".as_bytes();
let ptr = 0;
while let Some(b) = source.get(ptr) {
    if b == b'i' {
        if let Some(b2) = source.get(ptr + 1) {
            if b2 == b'd' {
              let id = unsafe { str::from_utf8_unchecked(ptr, ptr+1); }
            }
        }
    }
    ptr += 1;
}
```

(sorry for how stupid the code is). It would be insanely hard for the compiler to deduct that it can use boundary check elision, but we can deduct that quite easily - we just checked for what is at `ptr`, and what is at `ptr+1`.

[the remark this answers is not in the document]
Manishearth: why not store the string as `&str` and use `as_bytes()` in your processing code everywhere, so that you can safely subslice it with byte indices?

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
-//! AST, parser and serializer operations
```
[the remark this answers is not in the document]
zbraniecki: I think there's a value in keeping the parser in its own directory. This way I can easily tell which parts (ftlstream/errors) are parser specific, while ast stays crate specific (and in the future I expect serializer to get its own dir). I removed the errors dir.

zbraniecki:
> I would also (strongly) prefer to land the ResourceManager in a separate PR altogether. Let's make this PR about fluent-syntax, fluent-bundle and fluent-cli.

Done!

> (You might still want to clean up the commit history a bit, unless you're planning to squash all of it into a single commit).

I do plan to squash.

I'm keeping @Manishearth suggestion open, and I believe the rest of the feedback has been applied. I'll try to investigate his suggestion, but I believe this shouldn't block the rest of review and can be done in a separate PR prior to release.

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+            b'\n' => {
+                ps.ptr += 1;
+                return Ok((
+                    start_pos,
+                    ps.ptr,
+                    text_element_type,
+                    TextElementTermination::LineFeed,
+                ));
+            }
+            b'\r' if ps.is_byte_at(b'\n', ps.ptr + 1) => {
```
Pike: Can this go out-of-bounds for a stream ending in `\r`? No idea what happens in rust if that happens.

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+                true
+            } else {
+                false
+            }
+        }
+        _ => false,
+    };
+
+    if !is_valid {
+        //XXX: Generalize error type
```
Pike: Or the opposite, make the error type more specific?

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+                callee = ast::InlineExpression::FunctionReference { id };
+            }
+            let (positional, named) = get_call_args(ps)?;
+            ast::InlineExpression::CallExpression {
+                callee: Box::new(callee),
+                positional,
+                named,
+            }
+        } else {
+            return error!(ErrorKind::ForbiddenCallee, ps.ptr);
```
Pike: Does rust do early returns? I.e, could we just return an error on invalid, and not indent the valid part?

fluent-bundle/src/lib.rs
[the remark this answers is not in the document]
stasm: These are the three that we've seen used by the consumers of `fluent` in JS. I think it's OK to do it in a follow-up, too. Up to you.

fluent-syntax/src/parser/mod.rs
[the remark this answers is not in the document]
stasm: I still think that this is unnecessary proliferation of directories, but it's also a detail, so not blocking at all.

fluent-syntax/src/parser/ftlstream.rs
```diff
@@ -1,394 +1,211 @@
...
-        self.skip_inline_ws();
+        self.ptr += 1;
         Ok(())
     }
 
-    fn take_char_if(&mut self, ch: char) -> bool {
-        if self.ch == Some(ch) {
-            self.next();
-            return true;
+    pub fn take_if(&mut self, b: u8) -> bool {
```
stasm: Nit: Perhaps rename this to `take_byte` or `consume_byte` for symmetry with `expect_byte`?

fluent-syntax/src/parser/mod.rs
stasm: I think this is OK. `ParserStream::is_byte_at` uses the [checked `String::get` method](https://doc.rust-lang.org/std/string/struct.String.html#method.get) which returns `None` whenever the unchecked one would panic.

fluent-syntax/src/parser/ftlstream.rs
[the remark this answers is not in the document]
stasm: Digging a bit deeper into this, the panic seems to originate from `&self.source[start..end]` rather than the unsafe call to `from_utf8_unchecked`. Is there a checked way to index into a slice?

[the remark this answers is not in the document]
stasm: OK, so I think there are two things going on here. Can you please confirm that my understanding is correct?

1. When creating the slice of bytes with `&self.source[start..end]`, `end` might go out of bounds. We should check that it isn't or maybe use an iterator and the checked `slice::get` method. The `x = {"\\u` testcase exposes this bug.

2. This new slice is then passed to `from_utf8_unchecked` which is an unsafe method. If the slice of bytes contains a sequence of bytes which is invalid in UTF8, this method will panic. Realistically this is unlikely to happen with files which are read from disk and validated as UTF8, but might still be an issue when the data is not guaranteed to be UTF8.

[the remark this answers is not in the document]
stasm: > (…) might still be an issue when the data is not guaranteed to be UTF8.

Is this even possible given that ` FluentResource::try_new` takes a `String`?

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+            let slice_start = ps.ptr;
+            let mut indent = 0;
+            if text_element_role == TextElementPosition::LineStart {
+                indent = ps.skip_blank_inline();
+                if ps.ptr >= ps.length {
+                    break;
+                }
+                let b = ps.source[ps.ptr];
+                if indent == 0 {
+                    if b != b'\n' {
```
stasm: Nit: Do you think that using a different variable name than `b` here would be a good idea? I keep confusing it with the byte literal syntax, `b'x'`.

stasm: (You can also use a single `if` here.)

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+                        end,
+                        indent,
+                        text_element_role,
+                    ));
+                }
+            }
+
+            text_element_role = match termination_reason {
+                TextElementTermination::LineFeed => TextElementPosition::LineStart,
+                TextElementTermination::CarriageReturn => TextElementPosition::Continuation,
```
stasm: `CarriageReturn` should really be called `CRLF`, as we don't support the lone `CR` as a valid new line. Please add docs which explain why `TextElementTermination::CarriageReturn` results in something else than `TextElementTermination::LineFeed` does. In particular, AFAIUI, the quirk of the zero-copy design is that `a\r\n    b` becomes three `TextElements`: `a`, `\n` and `b`.

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+
+    if !has_default {
+        error!(ErrorKind::MissingDefaultVariant, ps.ptr)
+    } else {
+        Ok(variants)
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum TextElementTermination {
```
stasm: All the enums here need docs, which explain the _why_'s of this code. I suggested how to improve the docs of `CarriageReturn` in a comment below.

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+enum TextElementPosition {
+    InitialLineStart,
+    LineStart,
+    Continuation,
+}
+
+#[derive(Debug, PartialEq)]
+enum TextElementType {
+    Blank,
+    NonBlank,
```
stasm: I'm not entirely sure why `Blank` vs. `NonBlank` are needed. Please add a comment.

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+                TextElementTermination::PlaceableStart => TextElementPosition::Continuation,
+                TextElementTermination::EOF => TextElementPosition::Continuation,
+            };
+        }
+    }
+
+    if let Some(last_non_blank) = last_non_blank {
+        let elements = elements
+            .into_iter()
+            .take(last_non_blank + 1)
```
stasm: Why `+1`?

fluent-syntax/src/parser/mod.rs
stasm: Can `TextElementTermination` be combined with `TextElementPosition`?

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+                })
+            } else {
+                Ok(reference)
+            }
+        }
+        _ => Ok(reference),
+    }
+}
+
+fn get_literal<'p>(ps: &mut ParserStream<'p>) -> Result<ast::InlineExpression<'p>> {
```
stasm: There are only two literals in `Fluent` spec: `StringLiteral` and `NumberLiteral`. In `fluent.js` we called this function `getSimpleExpression`.

fluent-syntax/tests/fixtures/Makefile
```diff
@@ -0,0 +1,11 @@
+FTL_FIXTURES := $(wildcard *.ftl)
```
stasm: Remove this file please.

fluent-syntax/src/parser/mod.rs
stasm: (I understand that when serialized to JSON, these three `TextElements` become a single one, which is very neat.)

fluent-syntax/src/parser/ftlstream.rs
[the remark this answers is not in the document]
zbraniecki: > The x = {"\\u testcase exposes this bug.

I fixed it and reviewed other places where this may happen. Found no other so I think we're good.

> This new slice is then passed to from_utf8_unchecked which is an unsafe method. If the slice of bytes contains a sequence of bytes which is invalid in UTF8, this method will panic

No, we don't care if the file is UTF8 or if it contains invalid UTF8 tokens because the boundaries we set are validated to be. In other words, since Fluent tokens end on valid UTF8 chars like "\n", "{" etc. we can establish a range and know that we're slicing at the right place. What's inside, doesn't really matter to us. So, if the entry looks like this:

```
key = This is a XXX value { $foo }
```

then we slice out three times:
1) `key` which has a `start` and `end` known to be a valid alpha character and every character inside as well.
2) `This is a XXX value` which has `start` as the first non-space char, and `end` as one-before `{` char, so also no risk of cutting in the middle of a multi-byte char.
3) `foo` - similar to (1)

Even if in (2) the XXX is invalid UTF8, we don't care about it. It still slices properly because of what is a boundary in Fluent.

fluent-syntax/src/parser/mod.rs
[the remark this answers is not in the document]
zbraniecki: I prefer to keep it in its own directory. I'm experimenting with adding `lexer.rs` there as well and I like to keep parser specific code in it away from ast/serializer.

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+        }
+        ps.ptr = start;
+    }
+
+    let pattern = get_pattern(ps)?;
+
+    Ok(pattern.map(ast::Value::Pattern))
+}
+
+fn get_attributes<'p>(ps: &mut ParserStream<'p>) -> Result<Vec<ast::Attribute<'p>>> {
```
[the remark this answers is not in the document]
zbraniecki: For now, yes. I'd like to consider improving the recovery here eventually by recognizing that the error originated in the attribute, jumping to the next line that either is a start of new entry or starts with `__"."` and if the latter, resume parsing attributes for that entry.

This would allow us to recover from:

```
key = Value
    .attr = Value { $4 }
    .attr2 = Value 2
```

by salvaging the `attr2` into `key` message. But that's beyond the scope of this patch, so for now, I fixed it :)

fluent-syntax/src/parser/mod.rs
```diff
@@ -1,14 +1,842 @@
...
+    }
+}
+
+fn get_identifier<'p>(ps: &mut ParserStream<'p>) -> Result<ast::Identifier<'p>> {
+    let mut ptr = ps.ptr;
+
+    while let Some(b) = ps.source.get(ptr) {
+        if ps.is_byte_alphabetic(*b) {
+            ptr += 1;
+        } else if ptr == ps.ptr {
```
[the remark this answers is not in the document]
zbraniecki: Yes. Since `get_identifier` is such a hot loop I tested tons of variants and this is the fastest one. I'm going to switch to a more readable one for the sake of landing it sooner, and we can reevaluate perf optimizations later.

fluent-cli/src/main.rs
```diff
@@ -0,0 +1,149 @@
...
+fn print_entries_resource(res: &Resource) {
+    println!("{:#?}", res);
+}
+
+fn main() {
+    let matches = App::new("Fluent Parser")
+        .version("1.0")
+        .about("Parses FTL file into an AST")
+        .args_from_usage(
+            "-s, --silence 'disable output'
```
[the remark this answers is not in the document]
zbraniecki: I think I copied it from the JS impl and the goal was to be able to `time` the parser without timing the formatting and printing of the output. I changed it to only disable errors for now.

fluent-syntax/src/parser/ftlstream.rs
zbraniecki: `expect_*` is a fallible method that returns an error. I'll rename it to `take_byte_if`.

fluent-bundle/src/resolve.rs
```diff
@@ -191,100 +168,130 @@ impl ResolveValue for ast::Expression {
...
+
+impl<'source> ResolveValue for ast::InlineExpression<'source> {
+    fn to_value(&self, env: &Env) -> Result<FluentValue, ResolverError> {
+        match self {
+            ast::InlineExpression::StringLiteral { raw } => {
+                // XXX: We need to decode the raw into unicode here.
+                Ok(FluentValue::from(*raw))
+            }
+            ast::InlineExpression::NumberLiteral { value } => {
+                Ok(FluentValue::as_number(*value).unwrap())
```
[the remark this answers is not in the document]
zbraniecki: Not really. I must say that I see this patchset as bringing the parser to current state, and I expect us to only after that start cleaning up the resolver, so I was kind of ok with introducing half-baked unwraps like this. I'll file a follow-up for them.

fluent-syntax/src/parser/mod.rs
zbraniecki: Because of how `take` works. If `last_non_blank` is `0`, we need to take `1`.

zbraniecki:
Applied all feedback.


> OK, I think I'm mostly done with the review. It's a lot of code; thanks for the patience. I'd like to see a version of this PR with my comments addressed. The most important ones are:
> 
>     1. Comment the enums and their role in `get_pattern`.

Added.

>     2. Add protection against out-of-bounds indexing into `ps.source`. (And fix the bug I found yesterday.)

Removed the unsafe code for now, fixed the bug and verified that no other place uses `get_slice` with pointers that are not already checked to be in range.

>     3. Sometimes the code uses `unwrap` in places where I imagine panic are not possible if the AST was parsed by the parser. Please consider starting a conversation about the expectation towards this code when this assumption is not satisfied.

https://github.com/projectfluent/fluent-rs/issues/80


>     4. File issues for the things we know are follow-up material.

https://github.com/projectfluent/fluent-rs/issues/79

fluent-syntax/src/parser/mod.rs
stasm: Sure.

fluent-syntax/src/parser/mod.rs
zbraniecki: It keeps the track of which of the `elements` is the last one to be non_blank. Let's imagine this scenario:

```
key = This is
  multiline value

  and it still works


key2 = Value 2
```

For `key` it produces the following TextElements:

```
("This is\n", "   multiline value\n", "\n", "  and it still works\n", "\n", "\n")
```

So, we want to only take the lines `0..3`. It may be that since we filter out blank lines, it would be better to remove that variable and just filter out blank lines. I'll look into this after landing this :)

zbraniecki: https://github.com/projectfluent/fluent-rs/issues/81

fluent-syntax/src/parser/ftlstream.rs
```diff
@@ -186,7 +189,7 @@ impl<'p> ParserStream<'p> {
     }
 
     pub fn get_slice(&self, start: usize, end: usize) -> &'p str {
-        unsafe { str::from_utf8_unchecked(&self.source[start..end]) }
+        str::from_utf8(&self.source[start..end]).expect("Slicing the source failed")
```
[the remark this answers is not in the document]
zbraniecki: https://github.com/projectfluent/fluent-rs/issues/82

PR #86, opened 2019-01-31
Don't use `unwrap` in `ResolveValue`.

### LINKED ISSUE #80: Remove unwrap from Resolver
Resolver code uses some `unwrap` which should be replaced with error handling.

stasm:
Fix #80.

### DIFF
--- a/fluent-bundle/src/resolve.rs
+++ b/fluent-bundle/src/resolve.rs
@@ -205,7 +205,7 @@ impl<'source> ResolveValue for ast::InlineExpression<'source> {
 
                 for arg in named {
                     resolved_named_args
-                        .insert(arg.name.name.to_string(), arg.value.to_value(env).unwrap());
+                        .insert(arg.name.name.to_string(), arg.value.to_value(env).ok());
                 }
 
                 let func = match **callee {
--- a/fluent-bundle/src/resolve.rs
+++ b/fluent-bundle/src/resolve.rs
@@ -204,8 +204,10 @@ impl<'source> ResolveValue for ast::InlineExpression<'source> {
                 }
 
                 for arg in named {
-                    resolved_named_args
-                        .insert(arg.name.name.to_string(), arg.value.to_value(env).ok());
+                    if let Ok(arg_value) = arg.value.to_value(env) {
+                        resolved_named_args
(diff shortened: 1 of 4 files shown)


fluent-bundle/src/resolve.rs
```diff
@@ -154,9 +155,12 @@ impl<'source> ResolveValue for ast::Expression<'source> {
...
                             ast::VariantKey::NumberLiteral { value } => {
-                                let key = FluentValue::as_number(value).unwrap();
-                                if key.matches(env.bundle, selector) {
-                                    return variant.value.to_value(env);
+                                if let Ok(key) = FluentValue::as_number(value) {
+                                    if key.matches(env.bundle, selector) {
+                                        return variant.value.to_value(env);
+                                    }
+                                } else {
+                                    return Err(ResolverError::Value);
```
zbraniecki: `map_or_else`?

stasm: That's nightly-only, according to clippy. Also, we need the early return to exit the `for` loop but only when the `key` matches. Can you do that with a closure?

stasm:
> Can you elaborate why do you prefer to pass an unresolved argument to the function over not passing it at all?

No particular reason, I think both approaches are fine from the correctness perspective. Not inserting an unresolved named argument to the hashmap sounds like less boilerplate for writing functions, so let me add it. Thanks for the suggestion!

fluent-bundle/src/resolve.rs
```diff
@@ -200,8 +204,10 @@ impl<'source> ResolveValue for ast::InlineExpression<'source> {
                 }
 
                 for arg in named {
-                    resolved_named_args
-                        .insert(arg.name.name.to_string(), arg.value.to_value(env).unwrap());
+                    if let Ok(arg_value) = arg.value.to_value(env) {
+                        resolved_named_args
+                            .insert(arg.name.name.to_string(), arg_value);
+                    }
```
stasm: We might want to reconsider this code once `to_value` learns to report more than one resolution error.

fluent-bundle/src/resolve.rs
zbraniecki: nope, all good!

Issue #88, opened 2019-02-01
Simple ResourceManager


zbraniecki:
With 0.5 release, we reorganized our memory management story for resources, which now closer ties with the idea of using a ResourceManager to retrieve (and optionally cache) resources in memory.

I'd like to introduce a simple, file based resource manager for Fluent that can be used for simple projects.
This should lower the learning curve and help people kickstart their projects with less overhead. They would also likely want to use the example ResourceManager as a model for their own integration into their system.

I have a branch that adds a simple ResourceManager, but I'm also developing a more sophisticated one for Mozilla's needs and I'd like to consider bringing `fluent-resmgr` as a base for `l10nregistry-rs`.

zbraniecki:
landed today!

PR #89, opened 2019-02-01
Fluent Resource Manager

zbraniecki:
This is a WIP for #88.


fluent-resmgr/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "fluent"
```
bjorn3:
```suggestion
name = "fluent-resmgr"
```

PR #91, opened 2019-02-21
Zeronine bundle

zbraniecki:
This depends on #90 and updates the runtime to use 0.9 syntax.

### DIFF
--- a/fluent-bundle/Cargo.toml
+++ b/fluent-bundle/Cargo.toml
@@ -19,7 +19,7 @@ categories = ["localization", "internationalization"]
 
 [dependencies]
 fluent-locale = "^0.4.1"
-fluent-syntax = "^0.8"
+fluent-syntax = { path = "../fluent-syntax/" }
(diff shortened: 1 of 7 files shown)


fluent-bundle/Cargo.toml
```diff
@@ -19,7 +19,7 @@ categories = ["localization", "internationalization"]
 
 [dependencies]
 fluent-locale = "^0.4.1"
-fluent-syntax = "^0.8"
+fluent-syntax = { path = "../fluent-syntax/" }
```
stasm: Are you adding this just for testing? Or do you plan to keep this after this implementation is published?

zbraniecki: definitely will switch back after releasing fluent-syntax 0.9 and before releasing fluent-bundle 0.6 :)

zbraniecki:
sweet! My goal for 0.6 is to add more tests to bundle, so I'll use that!

Issue #87, opened 2019-02-01
Bring fluent-bundle test coverage above 90%


zbraniecki:
With 0.5 we got fluent-syntax to ~95% coverage! Now it's time to do the same to bundle

https://coveralls.io/github/projectfluent/fluent-rs?branch=master

zbraniecki:
Resolver tests:

- [x] Clean up the code
- [x] Improve errors
- [ ] Port resolver tests from js
  - [ ] arguments_test.js
  - [ ] attributes_test.js
  - [x] constructor_test.js
  - [x] context_test.js
  - [x] functions_runtime_test.js
  - [x] functions_test.js
  - [ ] literals_test.js
  - [ ] macros_test.js
  - [ ] patterns_test.js
  - [x] primitives_test.js
  - [x] select_expressions_test.js
  - [x] values_format_test.js
  - [x] values_ref_test.js
- [ ] Improve coverage

zbraniecki:
The coverage is now above 90%, but I'll wait with closing this issue until I land the rest of resolver tests from JS.

zbraniecki:
This is now completed!

Issue #74, opened 2018-12-13
Inconsistent error handling


JohnHeitmann:
If my bundle has the message:

```
foo = { bar }
```
Then `format('foo')` returns `('___', [])`. That alone was surprising. It seems like it's common behavior for most "stuff is missing" problems. Either `None` or `('___', [ warnings])` feel better to me. Middle layers should be able to learn and log about message glitches.

If my message was instead:
```
foo =
    -attr= something
```

Then `format('foo')` returns `('foo', [error])`. Those two kinds of errors feel like they should be handled the same way.

To get the best of graceful degradation, clear error signaling, and consistent error handling, a good return type might be `(string, [warnings])`. If you want to let higher layers skip a pointer chase for the NotFound case, then maybe instead of `[warnings]` you can use something like:

```
enum Warning {
    None,
    MessageNotFound,
    IncompleteMessage([details])
}
```

zbraniecki:
That's a good point! We'll want to remodel error handling soon, after we release the major revision to move us to zero-copy parsing and update the syntax to 0.8 (I have a branch that does it that I'm planning to stabilize soon!)

zbraniecki:
I landed a major rewrite that I think helps unify the error handling, but I'm not sure if we're done yet.

At the moment, we have four major "sources" of potential errors, and we handle them slightly differently:

* `fluent_syntax::parse` - returns `Result<ast::Resource, (ast::Resource, Vec<ParserError>)>`
* `fluent_bundle::resource::FluentResource::try_new` - returns `Result<Self, (Self, Vec<ParserError>)>`
* `fluent_bundle::FluentBundle::add_resource` returns `Result<(), Vec<FluentError>>`
* `fluent_bundle::FluentBundle::format` returns `Option<(Cow<'bundle, str>, Vec<FluentError>)>`

The way my thinking goes here is the following:

Fluent parser is lenient and attempts to recover as much as possible from the source. In result, it will *always* return something, but it *may* also return errors. I use `Result` here to indicate whether there are any errors, but I might as well just return a tuple and you can check if the second argument is empty for the same result.

Creating `FluentResource` using `try_new` is basically the same. Both return `ParserError`s.

`add_resource` is a bit different, because it only operates on the AST of `FluentResource` so basically whatever got rejected and ended up as errors from parser is ignored at this point (should be handled earlier).
[...]

PR #93, opened 2019-02-28
Refactor resolver errors to provide better fallbacking and return errors out of formatting

zbraniecki:
This is a pretty major refactor of the Resolver, in a pretty POC state of work.

It aims to achieve three main goals:

- Improve error handling, accumulate errors and report them to the user in a consistent manner (#74)
- Improve the test coverage by porting tests from the reference resolver to Rust (#87)
- Use `Cow<str>` to lower the CPU and memory cost of simple resolutions

The code is far from being completed, but it already renders really promising performance wins:

master:
```
Benchmarking resolve/"simple": Collecting 100 samples in estimated 5.0014                                                                         resolve/"simple"
                        time:   [423.91 ns 427.41 ns 431.49 ns]
                        change: [+307.52% +312.14% +317.44%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  1 (1.00%) high severe
Benchmarking resolve/"menubar": Collecting 100 samples in estimated 5.007                                                                         resolve/"menubar"
                        time:   [98.715 us 99.508 us 100.49 us]
                        change: [+188.03% +193.45% +199.59%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  2 (2.00%) high severe
Benchmarking resolve/"unescape": Collecting 100 samples in estimated 5.00                                                                         resolve/"unescape"
                        time:   [2.0391 us 2.0608 us 2.0865 us]
                        change: [+174.65% +179.37% +183.55%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

bundle-tests branch:
```
Benchmarking resolve/"simple": Collecting 100 samples in estimated 5.0002                                                                         resolve/"simple"
                        time:   [99.838 ns 100.40 ns 101.05 ns]
                        change: [-76.878% -76.609% -76.357%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Benchmarking resolve/"menubar": Collecting 100 samples in estimated 5.135                                                                         resolve/"menubar"
                        time:   [33.873 us 34.129 us 34.424 us]
                        change: [-66.651% -66.045% -65.508%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Benchmarking resolve/"unescape": Collecting 100 samples in estimated 5.00                                                                         resolve/"unescape"
                        time:   [735.93 ns 740.36 ns 745.43 ns]
                        change: [-65.005% -64.488% -63.949%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
```

All three benchmarks put the branch at 60-70% faster, mostly due to lack of string copying for simple strings.

I'd like to consider this refactor for 0.6.

@stasm - this is likely too early for review, but the POC is complete enough for you to asses the general approach. Lmk!


Pike:
Out of curiosity, have you considered the pre-evaluated algorithm we're using in the python resolver now?

The other thing we did in python was to move from dispatch to transformer pattern. That was merely because dispatch was slow. Is that a problem in rust?

Also, can we refer to the js resolver as the js resolver? We don't have a reference resolver yet ;-)

zbraniecki:
This now passes all tests, retains all functionality, and maintains the performance win of ~60% in resolver tests. I still want to:

[] Clean up the code
[] Improve errors
[] Port resolver tests from js
[] Improve coverage

but the code matures to a full PR now :)

zbraniecki:
@stasm - I'm still working on porting resolver tests and improving coverage, but I believe this code is ready for preliminary review.

If you prefer me to first document it, I'll postpone the review, but it may be easier for me to know if you approve the general direction before I dive into details.

zbraniecki:
I think that two interesting concepts that I introduced here, which may be useful to port back to JS are:

 - I moved the `tracking` to be more "lazy", instead of eager. That means that when you reference a message, I'll only add this message to the stack if the message goes deeper :) This actually gives a nice perf boost.
 - I introduced the `DisplayableNode` to unify where we turn nodes into displayable pieces and delay doing that until its needed.

zbraniecki:
@manishearth - I'd love to get your feedback on this PR. Majority of the real action is in src/resolve.rs

stasm:
> If you prefer me to first document it, I'll postpone the review, but it may be easier for me to know if you approve the general direction before I dive into details.

Thanks for letting me know. I'll be away next week, so it might best to postpone the review until I'm back.

zbraniecki:
Okay! I'll work on porting more tests, inline docs and getting review from Manish first then :)

fluent-bundle/src/resolve.rs
```diff
@@ -13,271 +13,294 @@ use std::hash::{Hash, Hasher};
...
+    Reference(String),
+    MissingDefault,
+    Argument(String),
     Value,
     Cyclic,
 }
 
 /// State for a single `ResolveValue::to_value` call.
-pub struct Env<'env> {
+pub struct Scope<'scope> {
```
Manishearth: This lifetime is the same as the one on the bundle, right?

We should give it a uniform name across the codebase and name it, perhaps something like `'bundle`?

Typically lifetimes are named the same as the source of their data, in this case it seems to be "computed Fluent information stored in the bundle". Unless I'm misreading it and `'scope` is a shorter lifetime.

It's also useful to document where the data behind the lifetime comes from so that it's easier to reason about this.

fluent-bundle/src/types.rs
```diff
@@ -11,41 +11,136 @@
...
+    Term,
+    Variable,
+    Function,
 }
 
-impl FluentValue {
-    pub fn into_number<S: ToString>(v: S) -> Result<Self, ParseFloatError> {
-        f64::from_str(&v.to_string()).map(|_| FluentValue::Number(v.to_string()))
+#[derive(Debug, PartialEq, Clone)]
+pub struct DisplayableNode<'v> {
```
Manishearth: Should all the `'v`s in this file be named `'source`?

fluent-bundle/src/resolve.rs
```diff
@@ -13,271 +13,294 @@ use std::hash::{Hash, Hasher};
...
 
-    fn track<F>(&self, identifier: &str, action: F) -> Result<FluentValue, ResolverError>
+    pub fn track<F>(&mut self, entry: DisplayableNode<'scope>, mut action: F) -> FluentValue<'scope>
     where
-        F: FnMut() -> Result<FluentValue, ResolverError>,
+        F: FnMut(&mut Scope<'scope>) -> FluentValue<'scope>,
     {
         let mut hasher = DefaultHasher::new();
-        identifier.hash(&mut hasher);
+        entry.id.hash(&mut hasher);
```
Manishearth: Should we just `(entry.id, entry.attribute).hash()`? That should do more or less the same thing. It _should_ produce the same hash, but I'm not 100% sure. Either way, we just need hash consistency here.

fluent-bundle/src/resolve.rs
```diff
@@ -13,271 +13,294 @@ use std::hash::{Hash, Hasher};
...
 
         if self.travelled.borrow().contains(&hash) {
-            Err(ResolverError::Cyclic)
+            self.errors.push(ResolverError::Cyclic);
+            FluentValue::Error(entry)
         } else {
             self.travelled.borrow_mut().push(hash);
-            self.scope(action)
+            let result = action(self);
+            self.travelled.borrow_mut().pop();
```
Manishearth: Could this instead be a HashSet with us removing the element when done? That seems faster on the `.contains()` call, plus we don't have to manually hash anything.

The only hazard with this is that cycles will reuse the same entry for the same element, however you forbid cycles above so it's fine.

The `Vec` isn't necessary here since we already have the call stack for keeping track of what got pushed last. A `Vec` would be useful if we decided to abandon recursion in favor of iteration, but if we're not doing that we should just use a HashSet and take the win on the `.contains()` call

zbraniecki: I tried to do that, and got a significant perf regression. Separated out to https://github.com/projectfluent/fluent-rs/issues/100

zbraniecki:
Gonna merge it, and continue working in separate issues.

Issue #104, opened 2019-04-17
Consider switching FluentSyntax::AST to use Cow instead of &str


zbraniecki:
AST in `FluentSyntax` currently uses `&str` which means that its basically read-only. As we mature the `fluent-syntax` crate we'll likely want to add a serializer and ability for users to operate on the AST.

There are two ways we could approach it:

1) We could add `OwnedMessage` which would be added to the `Resource::body` like any other node and a way to copy `Message` to a new `OwnedMessage`. This would be quite limiting, but likely sufficient for most use cases.

2) We could switch the AST to use `Cow<&str>` which means it'll still handle `&str` for all basic parsing/runtime work, but allow it to be replaced with owned `String` when modified/added.

(2) sounds tempting, but I'm not sure what are performance/memory consequences.

zbraniecki:
I have a PR now that pushes Cow into AST.

The performance impact is as follows:

parse/"simple": 19.8us -> 22us (+12.2%)
parse/"preferences": 310us -> 329us (+2.2%)
parse/"menubar": 70us -> 76us (+3.3%)

As expected the highest impact is on the simplest AST which would otherwise just use `&str` everywhere.

On the benefit side, having AST as `Cow` allows us to modify AST, and once we add Serializer get the whole `parse->modify->serialize` cycle required for most of the tooling.

Otherwise, we get a faster (lighter?) parser that is "read-only".

Thoughts on that? @manishearth, @stasm?

Manishearth:
> As expected the highest impact is on the simplest AST which would otherwise just use &str everywhere.

The AST isn't directly accessed by most consumers, yes? So it should be fine if it's a bit more complex.

zbraniecki:
Sorry, I misspoke. I meant that the simplest type of Fluent file (you can find our benches fixtures here: https://github.com/projectfluent/fluent-rs/tree/master/fluent-syntax/benches ) is the most affected by adding `Cow`.

I think it's an ok-tradeoff, but my question for you @Manishearth , was more about whether you think that it makes sense for the AST to use Cow to enable modifications of it. I'm not sure what other parser/serializer crates do.

zbraniecki:
I think we have all the data here, what's left is to make the decision.

Is writable AST worth the perf/mem penalty of `Cow`? Is it better to keep the AST/parser read-only and as fast as possible, and maybe introduce a separate AST for serialization with ability to easily convert read-only AST into read/write AST via some `From/Into` conversion?

[...]

PR #105, opened 2019-04-19
Switch AST to use Cow

zbraniecki:
Solves #104


zbraniecki:
Merged using AST over generic S.

Issue #82, opened 2019-01-24
Consider readding unsafe slicing in the parser


zbraniecki:
Thanks for removing the `unsafe` block here, at least for now. Would you mind filing a follow-up for discussing re-adding it, please? I'd like to look into it a bit more, given that we're likely protected from invalid UTF8 input by the sheer fact of accepting `String` into `FluentResource::try_new`.

_Originally posted by @stasm in https://github.com/projectfluent/fluent-rs/pull/76_

zbraniecki:
This change would give us ~20% performance win:

```
Benchmarking parse/"simple": Collecting 100 samples in estimated 5.0453 s (2                                                                            parse/"simple"          time:   [17.874 us 18.020 us 18.193 us]
                        change: [-20.946% -19.166% -17.259%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
Benchmarking parse/"menubar": Collecting 100 samples in estimated 5.6131 s (                                                                            parse/"menubar"         time:   [155.88 us 157.36 us 159.22 us]
                        change: [-21.585% -19.878% -18.093%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

Gnuplot not found, disabling plotting
```

zbraniecki:
Using latest Rust nightly and the PR that I just put against master, the perf impact is as following:

parse/"simple": -15%
parse/"preferences": -12.8%
parse/"menubar": -15.3%

zbraniecki:
@cyplo - on twitter you indicated you might be able to help us with fuzzing [0]. I may file a separate issue for the parser fuzzing prior to 1.0, but this is a particular aspect of it that I'd like to tackle now.

We have a parser that in theory should only cut at known character bounds. In result, we should be able to slice for free using unchecked, and we know that it brings sizable perf wins, but I don't know to verify the risk here.

Is that something you may know how to help us decide on?

[0] https://twitter.com/zbraniecki/status/1096383281929560064

cyplo:
Hi, sure thing @zbraniecki  ! Would it be possible to jump on a call to discuss this ? - you can DM me on Twitter

bjorn3:
[...]

PR #106, opened 2019-04-19
Use unsafe slicing

zbraniecki:
Solves #82

### DIFF
--- a/fluent-syntax/src/parser/ftlstream.rs
+++ b/fluent-syntax/src/parser/ftlstream.rs
@@ -148,7 +148,7 @@ impl<'p> ParserStream<'p> {
     }
 
     pub fn get_slice(&self, start: usize, end: usize) -> &'p str {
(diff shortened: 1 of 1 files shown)


fluent-syntax/src/parser/ftlstream.rs
```diff
@@ -148,7 +148,7 @@ impl<'p> ParserStream<'p> {
     }
 
     pub fn get_slice(&self, start: usize, end: usize) -> &'p str {
-        str::from_utf8(&self.source[start..end]).expect("Slicing the source failed")
+        unsafe { str::from_utf8_unchecked(&self.source[start..end]) }
```
bjorn3: This function should be unsafe. The start or end might not be at the start or end of a character, which makes the str invalid.

zbraniecki: @bjorn3 - please, comment in #82 where we're discussing it. In particular, I made a claim that the start and end must be at the start and end of a character due to how the parser works.

PR #109, opened 2019-06-23
Fix README.md

3c1u:
Added casts from `&'static str` into `String` to let the samples work. Also, on `fluent/README.md`, used the re-exports instead of using `fluent_bundle` directly.


zbraniecki:
Thank you!

PR #117, opened 2019-07-11
Switch to trashmap for scope tracking

### LINKED ISSUE #100: Consider replacing RefCell<Vec> with a HashSet in Scope
Could this instead be a HashSet with us removing the element when done? That seems faster on the `.contains()` call, plus we don't have to manually hash anything.

The only hazard with this is that cycles will reuse the same entry for the same element, however you forbid cycles above so it's fine.

The `Vec` isn't necessary here since we already have the call stack for keeping track of what got pushed last. A `Vec` would be useful if we decided to abandon recursion in favor of iteration, but if we're not doing that we should just use a HashSet and take the win on the `.contains()` call

_Originally posted by @Manishearth in https://github.com/projectfluent/fluent-rs/pull/93_

zbraniecki:
Fixes #100

### DIFF
--- a/fluent-bundle/Cargo.toml
+++ b/fluent-bundle/Cargo.toml
@@ -27,6 +27,7 @@ failure = "^0.1"
 failure_derive = "^0.1"
 intl_pluralrules = "^1.0"
 rental = "^0.5.4"
+trashmap = "^0.1.3"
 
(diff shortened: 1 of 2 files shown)


zbraniecki:
Abandoning.

Issue #119, opened 2019-07-12
Align FluentBundle with the new JS FluentBundle API


zbraniecki:
This is basically porting https://github.com/projectfluent/fluent.js/pull/380 to fluent-bundle-rs.

zbraniecki:
The initial performance is worrying.

```
Gnuplot not found, disabling plotting
resolve/"simple"        time:   [11.263 us 11.300 us 11.339 us]                              
                        change: [+19.138% +19.863% +20.589%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe
resolve/"preferences"   time:   [137.05 us 137.99 us 139.37 us]                                  
                        change: [+28.618% +30.002% +31.232%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high severe
resolve/"menubar"       time:   [44.473 us 44.652 us 44.834 us]                               
                        change: [+63.365% +64.480% +65.556%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
resolve/"unescape"      time:   [1.7313 us 1.7390 us 1.7468 us]                                
                        change: [-0.9290% -0.2554% +0.4500%] (p = 0.48 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

Gnuplot not found, disabling plotting
```

I haven't investigate the source of the regression yet. My initial guess is that creating two hashmaps (one to return from `get_message` and then the other to format) would likely be the culprit, but I'll investigate further.

zbraniecki:
Yep! Updating the benchmark to not use the `compound` wrapper, but rather resolve value/attributes just like the JS does yields a nice perf win compared to master:

```
Gnuplot not found, disabling plotting
resolve/"simple"        time:   [6.1629 us 6.1843 us 6.2070 us]                              
                        change: [-35.251% -34.824% -34.389%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low mild
[...]

PR #120, opened 2019-07-12
Update FluentBundle to the latest API

zbraniecki:
Implementes #119 119


zbraniecki:
I kept the convenience methods temporarily until I update the tests.

zbraniecki:
@stasm - I don't want to overload you as I know you're super-busy, so I'm not going to block on your review, but I think this patchset is ready and it puts us roughly on par with the JS.

There are still some areas missing (mainly lack of NUMBER/DATETIME and bomb protection), but I'd prefer to land this patch and continue catching up to JS functionality on master.

zbraniecki:
One change I placed here is to make l10n_args use owned strings as keys. `&str` as key sucks more often than not, and I filed #123 to come up with a clean solution that has better ergonomics (maybe Cow?). For now, owned keys should work without too much hassle.

fluent-bundle/examples/external_arguments.rs
```diff
@@ -20,23 +20,26 @@ unread-emails =
...
-    args.insert("name", FluentValue::from("John"));
+    args.insert("name".to_string(), FluentValue::from("John"));
 
-    match bundle.format("hello-world", Some(&args)) {
-        Some((value, _)) => println!("{}", value),
-        _ => println!("None"),
-    }
+    let msg = bundle.get_message("hello-world").expect("Message exists");
+    let mut errors = vec![];
+    let pattern = msg.value.expect("Message has a value");
```
stasm: `expect()` takes the description of the _error scenario_, rather than of the success one :)

fluent-bundle/examples/external_arguments.rs
```diff
@@ -20,23 +20,26 @@ unread-emails =
...
+    let value = bundle.format_pattern(&pattern, Some(&args), &mut errors);
```
stasm: It would be nicer to use `if let Some(value) = msg.value { .. }` guard around the call to `format_pattern` in all examples, because it's much closer to a real-world usage of the API.

fluent-bundle/src/bundle.rs
```diff
@@ -21,6 +21,12 @@ use crate::types::FluentValue;
 
 #[derive(Debug, PartialEq)]
 pub struct Message<'m> {
```
stasm: Would it make sense to call this `RawMessage`?

fluent-bundle/src/bundle.rs
```diff
@@ -43,10 +49,12 @@ pub struct Message<'m> {
...
 ///
 /// let mut args = HashMap::new();
-/// args.insert("name", FluentValue::from("Rustacean"));
+/// args.insert("name".to_string(), FluentValue::from("Rustacean"));
 ///
-/// let (value, _) = bundle.format("intro", Some(&args))
-///     .expect("Failed to format a message.");
+/// let msg = bundle.get_message("intro").expect("Message exists");
+/// let mut errors = vec![];
+/// let pattern = msg.value.expect("Message has a value");
```
stasm: Same as in examples, please use a proper `if let Some(value) = ...`.

fluent-bundle/src/bundle.rs
```diff
@@ -243,7 +251,44 @@ impl<R> FluentBundle<R> {
...
+        }
+        return Some(Message { value, attributes });
+    }
+
+    pub fn format_pattern<'bundle>(
+        &'bundle self,
+        pattern: &'bundle ast::Pattern,
+        args: Option<&'bundle HashMap<String, FluentValue>>,
+        errors: &mut Vec<FluentError>,
+    ) -> Cow<'bundle, str>
```
stasm: Are you still planning to change the return value of `format_pattern` to `(String, Option<Vec<Error>>)`?

fluent-bundle/src/bundle.rs
```diff
@@ -243,7 +251,44 @@ impl<R> FluentBundle<R> {
...
+        let mut attributes = if message.attributes.is_empty() {
+            HashMap::new()
+        } else {
+            HashMap::with_capacity(message.attributes.len())
+        };
+
+        for attr in message.attributes.iter() {
+            attributes.insert(attr.id.name, &attr.value);
+        }
+        return Some(Message { value, attributes });
```
stasm: Would it make more sense to do all this prep work in `add_resource` instead?

fluent-bundle/src/bundle.rs
stasm: Also, I can't comment on the exact line, but this docstring mentions `format` and `compound` a few times.

fluent-bundle/src/bundle.rs
zbraniecki: I'd prefer not to, at least for now.

fluent-bundle/src/bundle.rs
zbraniecki: Yeah, I'll look into this after I land this patch to verify the performance impact.

fluent-bundle/examples/external_arguments.rs
zbraniecki: I'm struggling with that a bit. I know this is a low level API that we expect to build on top of, but if someone would use it in their app, they wouldn't branch for a scenario where the string is missing, they'd expect the value of a string to be there if they placed it in their FTL file.

I don't want to suggest to localize an app by replacing:

```rust
println!("Enter your age:");
```

with:

```rust
if let Some(value) = msg.value {
  println!(bundle.format_pattern(value, None));
} else {
  unimplemented!()
}
```

stasm: >  if someone would use it in their app, they wouldn't branch for a scenario where the string is missing, they'd expect the value of a string to be there if they placed it in their FTL file.

This only works if _their FTL file_ is the only FTL file in the app. As soon as they have more FTLs for more locales, this assumption may not hold anymore, depending on who wrote them. As you know, Fluent is designed around the idea that translations must not break the app. Handling that `Option<Pattern>` is an important enforcement of this design.

I know this is an _example_, and it's OK to simplify. But there is a chance people will look at examples to make their first steps around Fluent, and I'd like to take the opportunity to educate :) There's also a chance people will copy&paste directly from examples.

Also, it's approximately the same amount of code, and some might argue that the `if let` is actually nicer.

```rust
let pattern = msg.value.expect("Message has no value");
let value = bundle.format_pattern(&pattern, None);
```

```rust
if let Some(value) = msg.value {
  println!(bundle.format_pattern(value, None));
}
```

Or, to make the null-value case explicit, I'd put a comment in the `else` block, like so:

```rust
if let Some(value) = msg.value {
  println!(bundle.format_pattern(value, None));
} else {
  // Show the fallback translation, e.g. the id of the message.
}
```

zbraniecki: > As you know, Fluent is designed around the idea that translations must not break the app.

Yes, but we achieve it on a higher level. We introduce nice fallback mechanisms that handle the case where the user expects a value and there is none.

On this level there is no such mechanism, no such fallback.

I believe for a reader of an example, seeing:

```rust
if let Some(value) = msg.value {
  println!(bundle.format_pattern(value, None));
}
```

is bizarre. It's basically saying "try to localize the message, and if that fails, show nothing".
I believe that `.expect()` is exactly the right call for the example. It says "Since you placed the value, expect it to be there, but please, remember that it's an `Option` and it may be null, so you'll likely have a macro/higher-level-api that will facilitate a fallback".

```rust
if let Some(value) = msg.value {
  println!(bundle.format_pattern(value, None));
} else {
  // Show the fallback translation, e.g. the id of the message.
}
```

This is more acceptable, but still kind of explicit. It gives the vibe of "Fluent makes your code way less readable.".

I'd prefer to use `.expect` to indicate that we expect the message, and I intend to improve docs to clearly state that the purpose of FluentBundle is to be handled by a higher level abstraction that will handle fallbacking, and since fallbacking may differ between bindings and apps, we make FluentBundle API clean for the higher level API to pick their strategy.

stasm: OK, that sounds fair, thanks!

fluent-bundle/src/bundle.rs
zbraniecki: I don't think so. We have `FluentBundle::get_message`, so we return a `Message`. There's no other `Message` in the crate.

stasm: how about `FluentMessage`, then? I'm concerned about the naming conflict with `ast::Message`. I understand that it's not really a problem in code, because the module namespacing works to our advantage. But conceptually, I'd prefer for these two message to have distinct names.

zbraniecki: Yep, filed an issue!

PR #130, opened 2019-07-27
Refactor fluent_args

### LINKED ISSUE #123: Refactor FluentArgs for better ergonomics
Currently FluentArgs is a HashMap over `&str` (and I'm going to change it to over `String`).

It would be nice to have it over `Cow` maybe and write some helper/macro to construct args.

zbraniecki:
Fixes #123


zbraniecki:
I went for a macro_rules. Would you suggest going for a proc macro instead?

PR #131, opened 2019-07-29
Add transform to FluentBundle

### LINKED ISSUE #97: Add pseudolocales
We should enable basic pseudolocales ported from JS

zbraniecki:
Fixes #97


zbraniecki:
This is just the scaffolding. I will later add more features from https://searchfox.org/mozilla-central/source/intl/l10n/L10nRegistry.jsm#494 like elongating, wrapping etc.

For now, I added the `transform` to `FluentBundle`, ported reference test for it, and added `fluent-pseudo` which exposes a simple `transform` function that can be fed into `FluentBundle::set_transform`.

@Manishearth - can you verify that I'm using Rust well here? I'd love to use type/trait alias for `TransformFn` (and for `FluentFunction`) so that people can just reference it when writing theirs, but it seems that those features are not yet available, to for now I have to manually put the signature and trait combination everywhere.

fluent-bundle/src/bundle.rs
```diff
@@ -238,6 +240,17 @@ impl<R> FluentBundle<R> {
         self.use_isolating = value;
     }
 
+    pub fn set_transform<F>(&mut self, func: Option<F>)
+    where
+        F: 'static + Fn(&str) -> String + Send + Sync
```
Manishearth: usually you write `'static` at the end

fluent-bundle/src/bundle.rs
```diff
@@ -238,6 +240,17 @@ impl<R> FluentBundle<R> {
         self.use_isolating = value;
     }
 
+    pub fn set_transform<F>(&mut self, func: Option<F>)
+    where
+        F: 'static + Fn(&str) -> Cow<str> + Send + Sync
+    {
+        if let Some(f) = func {
```
Manishearth: this whole thing can be written as `self.transform = func.map(Box::new)`

fluent-bundle/src/bundle.rs
```diff
@@ -335,9 +348,9 @@ impl<R> FluentBundle<R> {
     /// ```
     ///
     /// [FTL syntax guide]: https://projectfluent.org/fluent/guide/functions.html
-    pub fn add_function<F: 'static>(&mut self, id: &str, func: F) -> Result<(), FluentError>
+    pub fn add_function<F>(&mut self, id: &str, func: F) -> Result<(), FluentError>
     where
-        F: for<'a> Fn(&[FluentValue<'a>], &FluentArgs) -> FluentValue<'a> + Sync + Send,
+        F: 'static + for<'a> Fn(&[FluentValue<'a>], &FluentArgs) -> FluentValue<'a> + Sync + Send,
```
Manishearth:
```suggestion
        F:  for<'a> Fn(&[FluentValue<'a>], &FluentArgs) -> FluentValue<'a> + Sync + Send + 'static,
```

fluent-bundle/src/bundle.rs
```diff
@@ -238,6 +240,17 @@ impl<R> FluentBundle<R> {
         self.use_isolating = value;
     }
 
+    pub fn set_transform<F>(&mut self, func: Option<F>)
+    where
+        F: 'static + Fn(&str) -> Cow<str> + Send + Sync
```
Manishearth:
```suggestion
        F: Fn(&str) -> Cow<str> + Send + Sync + 'static
```

fluent-pseudo/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+]
+homepage = "http://www.projectfluent.org"
+license = "Apache-2.0/MIT"
+repository = "https://github.com/projectfluent/fluent-rs"
+readme = "README.md"
+keywords = ["localization", "l10n", "i18n", "intl", "internationalization"]
+categories = ["localization", "internationalization"]
+
+[dependencies]
+regex = "1.2"
```
Manishearth: newline at eof

fluent-bundle/src/bundle.rs
zbraniecki:
```rust
error[E0308]: mismatched types
   --> fluent-bundle/src/bundle.rs:247:26
    |
247 |         self.transform = func.map(Box::new)
    |                          ^^^^^^^^^^^^^^^^^^ expected trait std::ops::Fn, found type parameter
    |
    = note: expected type `std::option::Option<std::boxed::Box<(dyn for<'r> std::ops::Fn(&'r str) -> std::borrow::Cow<'r, str> + std::marker::Send + std::marker::Sync + 'static)>>`
               found type `std::option::Option<std::boxed::Box<F>>`
```

Pike:
Can we implement this in a way that the gecko wrapper can forward transform functions written in js?

This would be important for fluent-rs being a drop-in replacement for fluent.js.

zbraniecki:
We could try to produce an equivalent of `fluent-pseudo` in gecko and feed it to `set_transform` but it may be hard or impossible yet.

The rust side that this PR provides can be fed with anything so we can explore that

PR #144, opened 2019-12-11
Keep a running count of placables to protect against denial of service attacks

stasm:
A port of https://github.com/projectfluent/fluent.js/pull/439.

### DIFF
--- a/fluent-bundle/src/resolve.rs
+++ b/fluent-bundle/src/resolve.rs
@@ -18,13 +18,14 @@ use crate::resource::FluentResource;
 use crate::types::DisplayableNode;
 use crate::types::FluentValue;
 
-const MAX_PLACEABLE_LENGTH: usize = 2500;
+const MAX_PLACEABLES: u8 = 100;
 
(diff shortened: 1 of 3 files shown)


fluent-bundle/src/resolve.rs
```diff
@@ -35,6 +36,9 @@ pub struct Scope<'bundle, R: Borrow<FluentResource>> {
     args: Option<&'bundle FluentArgs<'bundle>>,
     /// Local args
     local_args: Option<FluentArgs<'bundle>>,
+    /// The running count of resolved placeables. Used to detect the Billion
+    /// Laughs and Quadrtic Blowup attacks.
+    placeables: usize,
```
zbraniecki: `u8` should be enough, no?

stasm: Yes. I briefly considered `u8` but went for `usize` since it was what `MAX_PLACEABLE_LENGTH` was using. I'll change both to `u8`.

PR #148, opened 2019-12-18
Intl memoizer

zbraniecki:


zbraniecki:
@Manishearth - here's what I got to. Not sure about the naming, but it mostly works with bumping/dropping refcounter.

The one thing that I failed to find how to do is how to remove all cached elements from all types when a rc for lang goes to 0. Can you help me with that?

zbraniecki:
Addresses #149

Manishearth:
@zbraniecki you can use a Weak instead of an Rc in the map itself. this will not remove map entries, but it will remove the actual data being cached

zbraniecki:
> @zbraniecki you can use a Weak instead of an Rc in the map itself. this will not remove map entries, but it will remove the actual data being cached

The way its meant to work is that the `IntlMemoizer` stores the entries between the first `get` and until the last `FluentBundle` of a given `lang` calls `drop_rc`.

If I use `Weak`, I'll need to have additional `Map` on each `FluentBundle` to store `Rc<T>` on it from the first time the bundle needs a formatter, until the bundle drops, right?

I'd prefer to only drop the entry in `drop_rc`. Is that possible?

Manishearth:
Hmm, the design constraints seem more complicated than i thought :/

zbraniecki:
@Manishearth - that's what we do I believe! Example:

```rust
       let mut memoizer = IntlMemoizer::default();
       let en_memoizer = memoizer.get_for_lang(lang);
       let mut en_memoizer_borrow = en_memoizer.borrow_mut();

       let cb = en_memoizer_borrow
            .try_get::<PluralRules>((PluralRuleType::CARDINAL,))
            .unwrap();
```

Is that different from what you suggested?

Manishearth:
Oh, huh, I feel like I read it the other way around at some point. r=me

PR #150, opened 2019-12-20
Intl memoizer use

zbraniecki:
@Manishearth - here's how I'd use the memoizer. One challenge I see is that there's no easy way to have a `Default` if I need to allocate an `IntlMemoizer` externally.

I'm wondering if it would be possible to have `FluentBundle` either take an external one or allocate its own if not provided, but I'm not sure how to design such a generic.

### DIFF
--- a/fluent-bundle/src/bundle.rs
+++ b/fluent-bundle/src/bundle.rs
@@ -6,12 +6,12 @@
 
 use std::borrow::Borrow;
 use std::borrow::Cow;
+use std::cell::RefCell;
 use std::collections::hash_map::{Entry as HashEntry, HashMap};
 use std::default::Default;
 
-use fluent_langneg::{negotiate_languages, NegotiationStrategy};
 use fluent_syntax::ast;
-use intl_pluralrules::{PluralRuleType, PluralRules};
+use intl_memoizer::IntlLangMemoizer;
 use unic_langid::LanguageIdentifier;
 
 use crate::entry::Entry;
@@ -127,7 +127,7 @@ pub struct FluentBundle<R> {
     pub locales: Vec<LanguageIdentifier>,
     pub(crate) resources: Vec<R>,
     pub(crate) entries: HashMap<String, Entry>,
-    pub(crate) plural_rules: PluralRules,
+    pub(crate) intls: RefCell<IntlLangMemoizer>,
     pub(crate) use_isolating: bool,
     pub(crate) transform: Option<Box<dyn Fn(&str) -> Cow<str> + Send + Sync>>,
 }
@@ -158,22 +158,15 @@ impl<R> FluentBundle<R> {
             .into_iter()
             .map(|s| s.clone().into())
             .collect::<Vec<_>>();
-        let default_langid = "en".parse().expect("Failed to parse `en` langid.");
-        let pr_locale = negotiate_languages(
-            &locales,
-            &PluralRules::get_locales(PluralRuleType::CARDINAL),
-            Some(&default_langid),
-            NegotiationStrategy::Lookup,
-        )[0]
-        .clone();
-
-        let pr = PluralRules::create(pr_locale, PluralRuleType::CARDINAL)
-            .expect("Failed to initialize PluralRules.");
+        let lang = locales
+            .get(0)
+            .expect("Expect at least one language")
+            .clone();
         FluentBundle {
             locales,
             resources: vec![],
             entries: HashMap::new(),
-            plural_rules: pr,
+            intls: RefCell::new(IntlLangMemoizer::new(lang)),
             use_isolating: true,
             transform: None,
         }
@@ -508,17 +501,13 @@ impl<R> FluentBundle<R> {
 
 impl<R> Default for FluentBundle<R> {
     fn default() -> Self {
-        let pr_langid: LanguageIdentifier = "en".parse().expect("Failed to parse `en` langid.");
         let langid = LanguageIdentifier::default();
-
-        let pr = PluralRules::create(pr_langid, PluralRuleType::CARDINAL)
-            .expect("Failed to initialize PluralRules.");
         FluentBundle {
-            plural_rules: pr,
-            locales: vec![langid],
+            locales: vec![langid.clone()],
             resources: vec![],
             entries: Default::default(),
             use_isolating: true,
+            intls: RefCell::new(IntlLangMemoizer::new(langid)),
             transform: None,
         }
     }
--- a/fluent-bundle/Cargo.toml
+++ b/fluent-bundle/Cargo.toml
@@ -26,6 +26,7 @@ intl_pluralrules = "5.0"
 rental = "0.5"
 smallvec = "1.0"
 unic-langid = "0.7"
+intl-memoizer = { path = "../intl-memoizer" }
 
 [dev-dependencies]
 criterion = "0.3"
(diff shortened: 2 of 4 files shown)


Manishearth:
Seems fine. I'd write a `FluentBundle::new_with_memoizer` function

fluent-bundle/src/bundle.rs
```diff
@@ -127,7 +130,7 @@ pub struct FluentBundle<R> {
     pub locales: Vec<LanguageIdentifier>,
     pub(crate) resources: Vec<R>,
     pub(crate) entries: HashMap<String, Entry>,
-    pub(crate) plural_rules: PluralRules,
+    pub(crate) intls: Rc<RefCell<IntlLangMemoizer>>,
```
Manishearth: you should make sure that going through a hashmap lookup (pretty cheap) each time you need the plural rules is not impacting perf too much

zbraniecki:
Hmm, continuing the process. Trying to hook one into another via a generic type to allow for construction of a `FluentBundle` that owns its `IntlLangMemoizer` vs one that borrows it.

Currently, I'm trying to get the bench to work, but when I do:

```rust
            let lang = langid!("en");
            let lids = &[lang.clone()];
            let mut memoizer = IntlMemoizer::default();
            let mut bundle = FluentBundle::new(lids, memoizer.get_for_lang(lang));
```

I'm getting:
```
error[E0277]: the trait bound `std::rc::Rc<std::cell::RefCell<intl_memoizer::IntlLangMemoizer>>: std::borrow::BorrowMut<intl_memoizer::IntlLangMemoizer>` is not satisfied
  --> fluent-bundle/benches/resolver.rs:90:54
   |
90 |             let mut bundle = FluentBundle::new(lids, memoizer.get_for_lang(lang));
   |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::borrow::BorrowMut<intl_memoizer::IntlLangMemoizer>` is not implemented for `std::rc::Rc<std::cell::RefCell<intl_memoizer::IntlLangMemoizer>>`
   |
   = note: required by `fluent_bundle::bundle::FluentBundle::<R, I>::new`

```

zbraniecki:
I think I'm messing up the `Rc<RefCell<IntlLangMemoizer>>` vs `RefCell<BorrowMut<IntlLangMemoizer>>` relationship. :(

zbraniecki:
Got it to work! Perf looks good I think:

```
▶ cargo bench
   Compiling fluent-bundle v0.9.0 (/projects/fluent-rs/fluent-bundle)
    Finished release [optimized] target(s) in 2.26s
     Running /projects/fluent-rs/target/release/deps/fluent_bundle-32f2072b4d2fe234

running 1 test
test types::tests::value_from_copy_ref ... ignored

test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out

     Running /projects/fluent-rs/target/release/deps/resolver-153afaede1425209
Gnuplot not found, disabling plotting
resolve/"simple"        time:   [5.2540 us 5.2693 us 5.2878 us]                              
                        change: [-0.6686% +0.3376% +1.0431%] (p = 0.51 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
resolve/"preferences"   time:   [47.641 us 47.726 us 47.818 us]                                   
                        change: [-4.8989% -4.5535% -4.2343%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe
resolve/"menubar"       time:   [13.357 us 13.379 us 13.405 us]                               
                        change: [-6.9899% -6.6010% -6.2043%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe
resolve/"unescape"      time:   [673.75 ns 675.38 ns 677.34 ns]                                
                        change: [-2.4000% -2.0749% -1.7354%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe

Gnuplot not found, disabling plotting
```

fluent-bundle/src/bundle.rs
```diff
@@ -158,27 +159,37 @@ impl<R> FluentBundle<R> {
...
-            Some(&default_langid),
-            NegotiationStrategy::Lookup,
-        )[0]
-        .clone();
+        let lang = locales.get(0).expect("Expect at least one language").clone();
+        FluentBundle {
+            locales,
+            resources: vec![],
+            entries: HashMap::new(),
+            intls: Rc::new(RefCell::new(IntlLangMemoizer::new(lang))),
```
stasm: I'm not sure if I like the fact that the lang memoizer is keyed on single language codes. _The_ reason why `FluentBundle::new` takes a vector of locales as argument is to be able to offer better negotiation results from intl formatters.

If keying on the lang is a necessity, perhaps consider parsing the lang code first, so that Unicode extensions are stripped, etc?

zbraniecki: > I'm not sure if I like the fact that the lang memoizer is keyed on single language codes. The reason why FluentBundle::new takes a vector of locales as argument is to be able to offer better negotiation results from intl formatters.

Correct, but in Rust case, we so far did not include negotiation as part of formatter constructor, but rather as an external operation.

In other words, we expect the locale provided to the constructor to be available, or the constructor will return an error.

This allows the callee to use any negotiation strategy they seem fit.

zbraniecki: We so far have not been enabling unicode extension keys to affect the formatters. This will change in the future with `unic-locale` replacing `unic-langid`.

stasm: > In other words, we expect the locale provided to the constructor to be available, or the constructor will return an error.

Do you mean the constructor for `FluentBundle`, or the formatter?

zbraniecki: FluentBundle. Basically, I'd expect there to be a moment where we do:

```rust
let pluralLocales = PluralRules::getAvailableLocales();
let datetimeLocales = DateTimeFormat::getAvailableLocales();
let numberLocales = NumberFormat::getAvailableLocales();
let fluentLocales = getL10nLocales();
let supportedLocales = negotiateLocales(dateTimeLocales, pluralLocales, numberLocales, fluentLocales);

let bundle = FluentBundle::create(supportedLocales);
```

stasm: Is there any reason left then to pass `supportedLocales` rather than the first `supportedLocale`? Should we change it?

zbraniecki: For now, I planned to do this negotiation from within of the constructor, but with the external formatters which should be considered for the negotiation, it may not be possible after all... so maybe it makes sense to just pass a single locale to the bundle?

stasm: @Pike Do you have more thoughts about this?

zbraniecki: May also be related - https://github.com/unicode-org/message-format-wg/issues/3#issuecomment-574777459 :)

zbraniecki:
I removed the shared memoizer constructor for now. Will revisit later.

Issue #146, opened 2019-12-16
Custom `FluentValue` types


zbraniecki:
At the moment, `fluent-bundle` accepts arguments to `format_pattern` that must be of type `FluentValue`.
`FluentValue` is an enum that can handle 4 different types - `Number`, `String`, `Error` and `None`.

Internally, `None` and `Error` are used in error scenarios, but from the variable point of view, we really only accept `String` and `Number`.
In JS world, we also accept `DateTime` type.

What's also important, is that in JS world we accept `Number` and we format it using `Intl.NumberFormat`, but in Rust case we can't really do that, so all that we get out of `Number` is that it matches and compares to FTL AST Numbers for selectors.

The final piece needed for https://bugzilla.mozilla.org/show_bug.cgi?id=1560038 is to hook `Intl.DateTimeFormat` and `Intl.NumberFormat` into `fluent-bundle-rs` and that requires changes to how we treat those `FluentValue` types.

In particular, I think we should investigate switching them to be traits, so that customers (in this case, Gecko) can implement their own traits that will provide formatting capabilities using `Intl` APIs.

@stasm - thoughts?

Pike:
I think traits make sense.

In the scenario of format-to-parts, stas and I also entertained the idea of having DOM nodes be `FluentValue`-wrapped things.

Can we design this trait such that it'd be usable if we had a format-to-parts API that was an iterator over `FluentValue`s?

stasm:
Agreed, a trait sounds like a good choice for this.

zbraniecki:
I talked to @Manishearth and he suggested that we don't switch the default types to `FluentType` trait because dynamic dispatching is quite costly.

Instead, we'll add a new FluentValue variant - `External(Box<dyn FluentTrait>)` and use it to add `DateTime` etc.

PR #153, opened 2020-01-07
Add FluentValue::Custom

zbraniecki:
Implements #146.

This is an implementation based on the dynamic dispatch approach suggested by @Manishearth .

There's an alternative approach using partial suggested by Lina - https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ad520cb4f7ece7a41e01a9079bdf7901

I like it more, but unfortunately it turns `FluentValue` into a generic and requires a pretty painful conversion of the whole codebase to `FluentValue<A>` in tons of places.

@Manishearth - lmk what you think, and if you prefer the generic partial approach I can go refactor the code for it.

### DIFF
--- a/fluent-bundle/src/types.rs
+++ b/fluent-bundle/src/types.rs
@@ -11,6 +11,7 @@
 //! [`resolve`]: ../resolve
 //! [`FluentBundle::format`]: ../bundle/struct.FluentBundle.html#method.format
 
+use std::any::Any;
 use std::borrow::{Borrow, Cow};
 use std::default::Default;
 use std::fmt;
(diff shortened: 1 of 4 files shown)


Manishearth:
> There's an alternative approach using partial suggested by Lina - https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ad520cb4f7ece7a41e01a9079bdf7901

If you're boxing the Custom one anyway there's not much additional benefit to doing that over dynamic dispatch, dynamic dispatch isn't much more expensive if you're boxing anyway.

fluent-bundle/src/types.rs
```diff
@@ -111,14 +111,60 @@ impl<'source> From<&ast::InlineExpression<'source>> for DisplayableNode<'source>
...
 
-#[derive(Debug, PartialEq, Clone)]
+pub trait FluentType: std::fmt::Debug {
+    fn duplicate(&self) -> Box<dyn FluentType>;
+    fn get_hash(&self) -> (&'static str, usize);
+    fn to_string(&self) -> Cow<'static, str>;
+    fn display(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result;
+}
+
+impl PartialEq for dyn FluentType {
```
Manishearth: A cleaner way to do this is

```rust
pub trait FluentType: AnyEq + 'static {
 // ...
}


impl PartialEq for dyn FluentType {
    fn eq(&self, other: &Self) -> bool {
        self.equals(other.as_any())
    }
}
pub trait AnyEq: Any + 'static {
    fn equals(&self, other: &dyn Any) -> bool;
    fn as_any(&self) -> &dyn Any;
}

impl<T: Any + PartialEq> AnyEq for T {
    fn equals(&self, other: &dyn Any) -> bool {
        if let Some(that) = other.downcast_ref::<Self>() {
            self == that
        } else {
            false
        }
    }
    fn as_any(&self) -> &dyn Any {
        self
    }
}

```
No need to have implementors write the hacky get_hash function.

fluent-bundle/src/types.rs
```diff
@@ -111,14 +111,60 @@ impl<'source> From<&ast::InlineExpression<'source>> for DisplayableNode<'source>
     }
 }
 
-#[derive(Debug, PartialEq, Clone)]
+pub trait FluentType: std::fmt::Debug {
+    fn duplicate(&self) -> Box<dyn FluentType>;
+    fn get_hash(&self) -> (&'static str, usize);
+    fn to_string(&self) -> Cow<'static, str>;
+    fn display(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result;
```
Manishearth: Why not just add a Display bound?

zbraniecki: I had to rename `to_string` to `as_string` in result of that to not override the Display's method, but I think that it works well.

fluent-bundle/src/types.rs
```diff
@@ -111,14 +112,69 @@ impl<'source> From<&ast::InlineExpression<'source>> for DisplayableNode<'source>
     }
 }
 
-#[derive(Debug, PartialEq, Clone)]
+pub trait FluentType: fmt::Debug + fmt::Display + AnyEq + 'static {
+    fn duplicate(&self) -> Box<dyn FluentType>;
```
Manishearth: Fwiw you can do the same thing here with an AnyClone trait

zbraniecki: How would a signature of that trait look like? `fn clone(&self) -> Box<dyn Any>;`? That feels off...

Manishearth: Hmm, fair. Yeah, this is trickier.

Issue #32, opened 2017-10-15
Add a lazy cache for Intl objects on MessageContext


zbraniecki:
Now that we have a proper `PluralRules` object, we should cache it on `MessageContext`.

zbraniecki:
Ehh, this ends up being a very non trivial task.

I spent way too much time today on #rust trying to understand the limitations. It boils down to  -  it's not easy/possible to have a single cache for different types. If all our formatters had the same set of methods, we could do this with `impl Trait` and `trait object`.
But in the world where we have `PluralRules.select` and `DateTimeFormat.format`, it's very hard.

It is possible to use an enum here, in a `closed` model which will not allow third-party devs to add their own formatters (or at least, they wont be cached).

Unfortunately, that requires us to change the mutability of `ctx` and `env` in resolver, because now any call can trigger adding an object to cache, so you have to pass `mut env` all around.

@stasm - would you be interested in taking a look at this? Maybe you'll come up with some better model for this problem, because I feel like I'm out of ideas :(

stasm:
We'll need some kind of mutability for `env` (via `mut` or `RefCell` maybe?) for #21, too. I'll first look into #21 and perhaps it will give us some ideas for this issue.

mmstick:
What needs to be cached, exactly?

zbraniecki:
nothing yet for we don't have any intl formatters. It's likely going to be delayed till later.

We're about to switch fluent to use https://github.com/unclenachoduh/pluralrules which will be the first intl formatter and it will not require any caching for now.

zbraniecki:
but once we get intl date time formatting and number formatting we'll likely want to cache the instantiated structures.

zbraniecki:
I was able to fix it trivially for now in https://github.com/projectfluent/fluent-rs/commit/7b0cd65c77b6011db9f2844996795b77f5f59bf5

Issue #31, opened 2017-10-15
Please dual-license


est31:
The gold standard for Rust projects is to dual license them under the Apache license and MIT. Most Rust crates are licensed that way. The Apache license gives users better patent protection, while the MIT is compatible to (L)GPL version 2 only projects. It would be nice if you would switch the license to dual license. The earlier the easier it is :).

Thanks!

zbraniecki:
That sounds good to me! @stasm ?

est31:
I see three contributors to the repo:

@zbraniecki 
@stasm 
@nvzqz

nvzqz:
Sounds good to me :+1:

stasm:
I'm personally fine with dual-licensing, but we should check with the [Licensing Team](https://www.mozilla.org/en-US/MPL/license-policy/). My understanding is that it should be OK. I filed [bug 1408992](https://bugzilla.mozilla.org/show_bug.cgi?id=1408992) to make sure.

stasm:
We've got a green light. I submitted #34.

stasm:
Thanks, everyone. We are officially dual-licensed now.

PR #162, opened 2020-02-25
Support maximum_fraction_digits, and fix a panic on 32 bit systems

dae:
Hi there,

Firstly, let me say thank you - I recently discovered Fluent, and it's a breath of fresh air after using gettext for a number of years. Thank you for all the thought and effort that has gone into it!

I had a user today report that our app was crashing when trying to render a translation with a floating point number. This turned out to be caused by an f64 with more than 9 decimal places - on a 32 bit system, it won't fit in a usize, so PluralOperands::try_from returns an error, and fluent-bundle .expect()s it not to.

The following patch adds support for maximum_fraction_digits, and caps it to 9 on 32 bit systems to ensure the usize won't overflow. I'm not sure if this is the best way to resolve the issue, but thought I'd share it in case it is useful.

### DIFF
--- a/fluent-bundle/src/types/number.rs
+++ b/fluent-bundle/src/types/number.rs
@@ -131,21 +131,31 @@ impl FluentNumber {
     }
 
     pub fn as_string(&self) -> Cow<'static, str> {
-        let mut val = self.value.to_string();
+        let mut max_frac_digits = self.options.maximum_fraction_digits.unwrap_or(15);
+        // since the plural code currently parses the resulting fractional digits into a usize, we can't
+        // have a precision above 9 digits on 32 bit platforms
+        if std::mem::size_of::<usize>() < 8 {
+            max_frac_digits = max_frac_digits.min(9);
+        }
+        // create the string with maximum precision
+        let with_max_precision = format!(
+            "{number:.precision$}",
+            number = self.value,
+            precision = max_frac_digits
+        );
+        // and then remove any excess trailing zeros
+        let mut val: Cow<str> = with_max_precision.trim_end_matches('0').into();
+        // adding back any required to meet minimum_fraction_digits
         if let Some(minfd) = self.options.minimum_fraction_digits {
-            if let Some(pos) = val.find('.') {
-                let frac_num = val.len() - pos - 1;
-                let missing = if frac_num > minfd {
-                    0
-                } else {
-                    minfd - frac_num
-                };
-                val = format!("{}{}", val, "0".repeat(missing));
-            } else {
-                val = format!("{}.{}", val, "0".repeat(minfd));
+            let pos = val.find('.').expect("expected . in formatted string");
+            let frac_num = val.len() - pos - 1;
+            let zeros_needed = minfd - frac_num;
+            if zeros_needed > 0 {
+                val = format!("{}{}", val, "0".repeat(zeros_needed)).into();
             }
         }
-        val.into()
+        // lop off any trailing '.', then return an owned value
+        val.trim_end_matches('.').to_string().into()
     }
 }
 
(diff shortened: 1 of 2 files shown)


zbraniecki:
Hi @dae! Thank you for kind words and the PR!

zbraniecki:
@dae - why did you decide to patch it in `fluent-bundle` rather than in https://github.com/zbraniecki/pluralrules/tree/master/intl_pluralrules ? It seems that, if I'm not mistaken, you're looking to prevent `TryInto<PluralOperands>` from panicking, no?

dae:
I first saw https://github.com/zbraniecki/pluralrules/blob/master/intl_pluralrules/src/operands.rs#L185 and thought about implementing a conversion from f64, but if my understanding is correct (and it may well not be!), I don't see how that could work, since intl_pluralrules won't be aware of the formatting options that are in use when displaying the string.

 I needed the ability to specify a maximum number of decimals anyway, so this was just the fastest way of addressing both issues. I guess https://github.com/zbraniecki/pluralrules/blob/master/intl_pluralrules/src/operands.rs#L93 could be adjusted to limit the number of digits instead.

Are you amenable to adding the maximum_fraction_digits part of the patch? If so, I can split this up and submit a separate request in intl_pluralrules for the capping part if that's what you'd prefer.

zbraniecki:
Hi, I dove into your patch and realized that you exposed a bug in the `intl_pluralrules` crate.

Fixing it here would make fluent not affected, but keep the original crate with that bug. So I went ahead and wrote a PR for pluralrules. I tested that it fixes it (with a minor change here) for fluent as well.

Let me know if that solution is acceptable for you!

dae:
I tried testing it, but I presume you haven't pushed the changes required to this repo yet:

```
error[E0308]: mismatched types
   --> fluent-bundle/src/types/number.rs:225:31
    |
225 |                 operands.f *= 10_usize.pow(mfd as u32 - operands.v as u32);
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u64`, found `usize`

error[E0277]: cannot multiply-assign `usize` to `u64`
   --> fluent-bundle/src/types/number.rs:225:28
    |
225 |                 operands.f *= 10_usize.pow(mfd as u32 - operands.v as u32);
    |                            ^^ no implementation for `u64 *= usize`
    |
    = help: the trait `std::ops::MulAssign<usize>` is not implemented for `u64`
```

If it addresses the panics then that's great :-)

Re issues #32 and #31 mentioned on #33, I fear I am still missing something - don't the plural rules depend on how the number is to be formatted? For example if you have a float 1.0, if maximum decimal places were 2 then it would be "1.00 apples", but if they were 0, it would be "1 apple". So don't you either need to pass float in as a pre-formatted string, or make pluralrules aware of how the number is going to be formatted at the same time?

zbraniecki:
> I tried testing it, but I presume you haven't pushed the changes required to this repo yet:

You just need to change `10_usize` to `10_u64`.

zbraniecki:
> So don't you either need to pass float in as a pre-formatted string, or make pluralrules aware of how the number is going to be formatted at the same time?

You may be right! I may have to extend it there *^^*!

dae:
Thanks, I can confirm the my previous test no longer panics when using the updated pluralrules.

The thing that struck me as potentially difficult about starting with a float is that when the decimal places haven't been rigidly set by the caller, you'd need to determine how many decimal places are appropriate, and it seems like you'd end up reimplementing the logic in format that does so

```
assert_eq!(&format!("{}", 3.020000000000000007), "3.02");
```

The logic looks to be non-trivial:

https://github.com/rust-lang/rust/blob/master/src/libcore/num/flt2dec/mod.rs

zbraniecki:
Yes, I agree. It's just a separate issue for me from a crash :)

dae:
Closing this, as the panic issue has been fixed, and it's possible to enforce a maximum digit limit in a custom formatter instead.

PR #184, opened 2020-06-23
Create a FTL serializer

### LINKED ISSUE #182: FTL Serializer
At work we're planning to use Google Translate to generate some initial translations for our app. Does a pretty-printer already exist to help turn the translated `FluentMessage`s back into text form?

If not, I'll put together my own and publish it to crates.io.

Michael-F-Bryan:
Fixes #182.

### DIFF
--- a/fluent-syntax/src/lib.rs
+++ b/fluent-syntax/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod ast;
 pub mod parser;
 pub mod unicode;
+pub mod serializer;
 
 #[cfg(feature = "json")]
 pub mod json;
--- a/fluent-syntax/src/serializer.rs
+++ b/fluent-syntax/src/serializer.rs
@@ -1,0 +1,343 @@
+use crate::ast::*;
+use std::io::{Error, Write};
+
+pub fn serialize(resource: &Resource<'_>) -> String {
+    let mut buffer = Vec::new();
+    let options = Options::default();
+
+    Serializer::new(&mut buffer, options)
+        .serialize_resource(resource)
+        .expect("Writing to an in-memory buffer never fails");
+
+    String::from_utf8(buffer).expect("The serializer only ever emits valid UTF-8")
+}
+
+#[derive(Debug)]
+pub struct Serializer<W> {
+    writer: W,
+    options: Options,
+    state: State,
+}
+
+impl<W: Write> Serializer<W> {
+    pub fn new(writer: W, options: Options) -> Self {
+        Serializer {
+            writer,
+            options,
+            state: State::default(),
+        }
+    }
+
+    pub fn serialize_resource(&mut self, res: &Resource<'_>) -> Result<(), Error> {
+        for entry in &res.body {
+            match entry {
+                ResourceEntry::Entry(entry) => self.serialize_entry(entry)?,
+                ResourceEntry::Junk(junk) if self.options.with_junk => self.serialize_junk(junk)?,
+                ResourceEntry::Junk(_) => continue,
+            }
+
+            self.state.has_entries = true;
+        }
+
+        Ok(())
+    }
+
+    fn serialize_entry(&mut self, entry: &Entry<'_>) -> Result<(), Error> {
+        match entry {
+            Entry::Message(msg) => self.serialize_message(msg),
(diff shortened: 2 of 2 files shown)


Michael-F-Bryan:
I've added all tests down to ["nested select expression"](https://github.com/projectfluent/fluent.js/blob/6f9eec101cbdf3bff1fa398ce4cd7ab213d8fc31/fluent-syntax/test/serializer_test.js#L354) and they all pass locally. 

At this point there are no more `unimplemented!()`s so I believe I'm hitting every branch, now I just need to make sure this is identical to the TypeScript implementation. 

@zbraniecki, would you be able to nominate someone as a reviewer? I'm going to finish up for today, but this should reach parity with the TypeScript serializer by tomorrow morning.

zbraniecki:
> @zbraniecki, would you be able to nominate someone as a reviewer? I'm going to finish up for today, but this should reach parity with the TypeScript serializer by tomorrow morning.

Looks good! I'll try to review by EOW.

Michael-F-Bryan:
Thanks @zbraniecki! We'll use git dependencies for now, so that takes the pressure off you.

I've implemented all the TypeScript tests and they pass, so this should be identical to the TypeScript implementation. Having an existing serializer I can copy from, complete with a thorough test suite, made this job really easy 🙂

Pike:
Watch out for https://github.com/projectfluent/fluent.js/issues/512

Michael-F-Bryan:
@zbraniecki, any chance you can review this?

zbraniecki:
> @zbraniecki, any chance you can review this?

I will. I'm sorry for the delay. As you may have heard, we're going through a bit of a change and I'm per-occupied with some non-technical tasks, but `fluent-rs` is maintained and I do intend to get another release soonish.

Please, be patient and feel free to use a fork in the meantime, it shouldn't bitrot.

SirWindfield:
On a side note, wouldn't it be more beneficial to use `serde` and leverage its powers? You don't have to use the `derive` macros that `serde` provides but can implement the traits on your own. The real benefit comes from being able to serialize to other file formats than text. 

In such a case, an application could theoretically provide its l10n files in a binary format through Over-the-air updates, reducing the number of bytes to transfer compared to using a text representation. Deserializing using `serde` would allow users to also use binary formats, reducing the overall file size. It's probably not much, but on smaller systems like embedded devices, a few bytes are already worth it.

Michael-F-Bryan:
@SirWindfield have you seen [the `fluent_syntax::json` module](https://github.com/projectfluent/fluent-rs/blob/961450bbbfc2b20e9920b5aa1d5384a2dda3922e/fluent-syntax/src/json/mod.rs)?

The main reason to also have a manual serializer is because the `serde` serializer API isn't really compatible with a pretty-printer.

`serde` works by encoding a state machine that can traverse a JSON-esque tree of data. Some big assumptions are that how something is serialized won't change based on its value or the value of a child-node and that you're just dealing with atoms (strings, numbers, etc.), sequences (lists), and maps (maps and structs).

On the other hand, how you print parts of a `fluent_syntax` AST depends very much upon their neighbours or their child nodes. You  may need to [insert whitespace](https://github.com/projectfluent/fluent-rs/blob/34f7a939801bcb1035b1683f3c10337b88bc6e44/fluent-syntax/src/serializer.rs#L128-L148) and [special-case some nodes](https://github.com/projectfluent/fluent-rs/pull/184/files#diff-b96ac4bedc98b9bcb819fa6d1b773538R184-R189) to make sure the final result looks correct to a human. 

> In such a case, an application could theoretically provide its l10n files in a binary format through Over-the-air updates, reducing the number of bytes to transfer compared to using a text representation. 

I'd just compress the files. The vast majority of the AST is plaintext that wouldn't disappear after being parsed, so you don't gain much by using a binary format.

zbraniecki:
I think this would be great now that we have AST over S. You could make it work for any `Slice<ToString>`, and this way any AST will be serializable!

I'm planning to generalize Resolver soon to work on any `Slice<AsRef<str>>`.

zbraniecki:
Hi @Michael-F-Bryan . I'm going to close this PR in favor of a follow up in #241.
If you think we should keep it open, let me know

PR #185, opened 2020-06-24
Implemented Clone for AST nodes

Michael-F-Bryan:
When implementing the code for rewriting parts of an AST, I noticed none of the AST types implement `Clone`.

Not having the impl means f you don't need to update a node and just want a copy of it, you have to implement the clone logic yourself instead of just writing `node.clone()`,

Is there any major reason why we don't `#[derive(Clone)]`?


zbraniecki:
I incorporated it into https://github.com/projectfluent/fluent-rs/pull/198

PR #186, opened 2020-06-25
Implement std::error::Error for ParserError

Michael-F-Bryan:
Implementing `std::error::Error` and `Display` for `ParserError` can make handling parser errors and notifying the user a bit easier. I had to [implement my own](https://gitlab.com/wintech-engineering/fluent-machine-translations/-/blob/2cb4bdb3d035a855a5c0c1c16da20a388e523a83/cli/src/main.rs#L125-208) logic for this, and matching on each `ErrorKind` variant can be annoying.

I'm not sure if we want the `Display` impl to just show the reason parsing failed (i.e. defer to the `self.kind` impl), or whether we also want to mention `pos` in the error message. I believe `pos` represents the index into the string and not line/column numbers, so it's not very useful for a human wanting to fix the parse error.

### DIFF
--- a/fluent-syntax/src/parser/errors.rs
+++ b/fluent-syntax/src/parser/errors.rs
@@ -1,8 +1,18 @@
+use std::fmt::{self, Display, Formatter};
+
 #[derive(Debug, PartialEq)]
 pub struct ParserError {
     pub pos: (usize, usize),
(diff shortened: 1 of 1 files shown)


zbraniecki:
that looks good, thank you!

PR #187, opened 2020-07-05
Reference-counted AST

Michael-F-Bryan:
This is an alternative to #183 which uses a reference-counted string instead of copy-on-write.


zbraniecki:
@Michael-F-Bryan does https://github.com/projectfluent/fluent-rs/pull/198 address your goal?

zbraniecki:
I'm going to close it since generic AST landed. reopen if needed.

PR #188, opened 2020-07-16
Allow AST traversal via the Visitor trait

Michael-F-Bryan:
See https://github.com/projectfluent/fluent-rs/pull/183#issuecomment-651517745

### DIFF
--- a/fluent-bundle/src/bundle.rs
+++ b/fluent-bundle/src/bundle.rs
@@ -10,6 +10,7 @@ use std::collections::hash_map::{Entry as HashEntry, HashMap};
 use std::default::Default;
 
 use fluent_syntax::ast;
+use fluent_syntax::visit::Visitor;
 use unic_langid::LanguageIdentifier;
 
 use crate::entry::Entry;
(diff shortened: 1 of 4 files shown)


Michael-F-Bryan:
As mentioned in https://github.com/projectfluent/fluent-rs/pull/183#issuecomment-659661542:

> It's possible to build a new tree by using a pushdown automaton which will build up a new AST on each transition, but you'll likely run into issues with the borrow checker.... When you recurse and add a new state to the stack, you'd be forced to "find" where you want to insert new nodes instead of just holding a reference to the parent in the automaton (multiple nested states means multiple mutable pointers into the same tree, which is no bueno). 
>
> I'm sure you could work around this with enough dynamic dispatch and indirection, but to be honest I'd just [write the traversal/modification by hand](https://gitlab.com/wintech-engineering/fluent-machine-translations/-/blob/d10741e14f33ecdaa7d57bf97da269f267df3b6d/lib/src/sub.rs#L57-148) at that point. That way you're free to do whatever you want and don't need to conform to a particular interface.

This experiment was useful, but I don't think we'll want to go down this path so I'm going to close the PR.

PR #194, opened 2020-09-15
Introduce Writer to Resolver

zbraniecki:
This PR separates the `WriteValue` from `ResolveValue` and enables writing to a predefined buffer for most cases, except when we need to resolve a value for Selector.


zbraniecki:
Superseeded by https://github.com/projectfluent/fluent-rs/pull/199

PR #202, opened 2020-11-06
Improve readability of accented pseudolocale

dminor:
This only applies accents to consonants to improve readability. It also adds
elongation for the letter 'i' which was missed previously.


dminor:
Or perhaps 'i' was left out intentionally, in which case I'll add a comment :)

zbraniecki (approved):

PR #205, opened 2020-11-11
Re-instate Debug trait for FluentArgs

### LINKED ISSUE #203: Debug support regression for FluentArgs
In 0.12.0, FluentArgs used to support Debug, so I could include it in other structs and `#[derive(Debug)]` on them and get meaningful output. That support seems to have disappeared in 0.13.1 and I'm having trouble upgrading apps as a result :-/

alerque:
Closes #203


zbraniecki (approved):
ty!

PR #206, opened 2020-11-12
Fix test expectations for test_pseudo

dminor:
The behaviour changed as part of https://github.com/projectfluent/fluent-rs/pull/202
but this test was not updated.


zbraniecki (approved):
ty!

PR #208, opened 2020-11-24
Add License Files to Crates

fhoehle:
This PR adds the two possible licenses file to all published crates. This would ease to compile a list of all licenses and corresponding files for packages which depend on `projectfluent` rust crates. `cargo` does not support mutiple license files therefore links are used see related PR for `cargo` (https://github.com/rust-lang/cargo/issues/3537). I would be glad to receive some feedback and/or review.


zbraniecki (approved):
thank you!

PR #209, opened 2020-12-14
clarify the purpose of the locales argument

dae:
I found the previous documentation confusing, and was passing the
user's full language preferences in, which lead to languages using the
wrong plural rules in the case where the first language preference was
missing strings.


zbraniecki (approved):

zbraniecki:
thank you! Can you resolve the conflict?

dae:
Done

zbraniecki:
thank you!

PR #215, opened 2021-01-19
Fix useGrouping in fluent-bundle

dminor:
The `useGrouping` number option is assigned a default value, but never updated.
In the Intl.NumberFormat API, `useGrouping` is a boolean. Since Fluent does not
support boolean values, a string is used here instead. If the value is `"false"`,
grouping is disabled, otherwise it is enabled.

This was missed because the unit tests tested for `"true"`, which is the default
value for useGrouping.


zbraniecki:
thank you!

zbraniecki (approved):
lgtm!

Issue #213, opened 2021-01-12
Unable to parse multiline messages formatted with CRLF with breaks that don't match the indentation of the rest of the message


kellpossible:
When I have a message like:

```txt
multiline-message = 
    This is a multiline message 1.

    This is a multiline message 2.
```

Fluent successfully parses when the file is formatted with LF, but has a parse error when the file is formatted with CRLF.

However if the gap between the lines contains indentation matching the other lines, then it successfully parses with both LF and CRLF.

```txt
multiline-message = 
    This is a multiline message 1.
    
    This is a multiline message 2.
```

Editors often automatically remove this indentation, or don't place it in the first place.

kellpossible:
I have created a minimal reproduction of the issue in the [`multiline-break-indentation-bug` branch of this repo](https://github.com/kellpossible/fluent-multiline-bug/tree/multiline-break-indentation-bug)

Pike:
Fly-by note, the ´crlf´ test is disabled, https://github.com/projectfluent/fluent-rs/blob/cd8d6c4e7a20ff69762ff86c7a169df790d50ddd/fluent-syntax/tests/parser_fixtures.rs#L21-L23

Did you test your example on other implementations by any chance?

Pike:
Also, can you set `.gitattributes` on your example repo to ensure that we all look at the same line endings?

kellpossible:
> Also, can you set `.gitattributes` on your example repo to ensure that we all look at the same line endings?

I'm not familiar with `.gitattributes` but I gave it a try, please let me know if you think it's correct

Pike:
Yes, thanks. Sadly not through the gh UI, but when cloning it does.

Also tested the python ipml just now, that parses OK

zbraniecki:
This should be now fixed by https://github.com/projectfluent/fluent-rs/commit/0195099c8a71dcfa2acd05578e94ed6d1203549a - can someone verify?

zbraniecki:
Closing. Reopen if needed.

PR #217, opened 2021-01-22
Mock test for issue #213

Pike:
This isn't a sound way to hook up this test, as it taints the upstream reference fixtures.

But I guess it'd be good way to give #213 a boost.

FWIW, I ran the `crlf.ftl` test locally, and it produced the expected results, though the tests fail.

This test fails in an unexpected way, in that it's not the `TextElement` that's segmented differently, but that the continuation isn't detected.

### DIFF
--- a/fluent-syntax/.gitattributes
+++ b/fluent-syntax/.gitattributes
@@ -1,2 +1,3 @@
 tests/fixtures/crlf.ftl eol=crlf
+tests/fixtures/crlf-issue213.ftl eol=crlf
 tests/fixtures/cr.ftl eol=cr
--- a/fluent-syntax/tests/fixtures/crlf-issue213.ftl
+++ b/fluent-syntax/tests/fixtures/crlf-issue213.ftl
@@ -1,0 +1,8 @@
+multiline-message-indented-breaks = 
+    This is a multiline message 1.
+    
+    This is a multiline message 2.
+multiline-message = 
+    This is a multiline message 1.
+
+    This is a multiline message 2.
--- a/fluent-syntax/tests/fixtures/crlf-issue213.json
+++ b/fluent-syntax/tests/fixtures/crlf-issue213.json
@@ -1,0 +1,65 @@
+{
+    "type": "Resource",
+    "body": [
+        {
+            "type": "Message",
+            "id": {
+                "type": "Identifier",
+                "name": "multiline-message-indented-breaks"
+            },
+            "value": {
+                "type": "Pattern",
+                "elements": [
+                    {
+                        "type": "TextElement",
+                        "value": "This is a multiline message 1."
+                    },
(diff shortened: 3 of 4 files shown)


zbraniecki:
Thanks to the test case, the fix was easy!

```diff
diff --git a/fluent-syntax/src/parser/pattern.rs b/fluent-syntax/src/parser/pattern.rs
index 84804b0..47b801b 100644
--- a/fluent-syntax/src/parser/pattern.rs
+++ b/fluent-syntax/src/parser/pattern.rs
@@ -75,7 +75,7 @@ where
                     }
                     let b = self.source.as_ref().as_bytes().get(self.ptr);
                     if indent == 0 {
-                        if b != Some(&b'\n') {
+                        if b != Some(&b'\r') && b != Some(&b'\n') {
                             break;
                         }
                     } else if !Self::is_byte_pattern_continuation(*b.unwrap()) {
```

Can you add this to your PR and we can merge it with the test?

Pike:
I've filed an issue upstream. As per my comment, I think just putting additional fixtures in the rust repo isn't great.

Regarding your proposed fix, I wonder how that fares with pure `cr` mac-like files, so I extended the expected test change to do both.

As for the rust impl, I wonder if we should just hard-patch the fixture on the json side to enable the `crlf` test? That's still tainting the fixtures, but maybe in a more straight-forward way?

zbraniecki:
> Regarding your proposed fix, I wonder how that fares with pure cr mac-like files, so I extended the expected test change to do both.

You mean, just to prevent crashes, right? We don't support `CR` line ends: https://github.com/projectfluent/fluent/blob/master/spec/fluent.ebnf#L136

> As for the rust impl, I wonder if we should just hard-patch the fixture on the json side to enable the crlf test? That's still tainting the fixtures, but maybe in a more straight-forward way?

I fixed that in our tests API just now. Which means that you can either make the test split `\n` or combine them the way reference fixtures do, and it'll work.

zbraniecki:
Can you include my fix and rebase on top of master? This should be landable then.

zbraniecki:
Oh, and with my fix, you can just extend the `crlf.ftl` file instead of introducing a new one! And upstream it :)

Pike:
> 
> 
> Oh, and with my fix, you can just extend the `crlf.ftl` file instead of introducing a new one! And upstream it :)

Right, that was where I was intending to go.

zbraniecki:
> Right, that was where I was intending to go.

Ok! In that case my proposal is:

1) I land my fix here
2) We land the fixture to test for it upstream
3) I vendor it in back here

How does it sound?

zbraniecki:
@Pike can we close this?

Pike:
I guess so. I left you a needinfo-ish thing upstream.

PR #218, opened 2021-01-24
Fix badges

Pike:
The badges in the readme are still pointing to travis.

Fixing https for meritbadge as a fly-by, to make the badges work in vscode preview. Also dropping the badge for `fluent-cli` as that's not on crates.io


zbraniecki (approved):
lgtm!

PR #221, opened 2021-03-13
Swap ouroboros for once_self_cell

zbraniecki:
There's a new self-referential crate in town aiming for exactly the use case we need and claiming lower build time and vendor-in overhead - https://github.com/Voultapher/once_self_cell


zbraniecki:
This is still a draft because we're waiting for https://github.com/Voultapher/once_self_cell/issues/1

zbraniecki:
Fixed by #227.

PR #224, opened 2021-05-02
fluent-fallback-cache: Fix waiting from multiple tasks.

emilio:
The cache effectively multiplexes a stream, so we can't just rely on the
stream receiving end to wake up all the relevant tasks.

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1613705#c56 for
me.

Disclaimer: I'm not the most familiar with async rust, so take with a
grain of salt.

### DIFF
--- a/fluent-fallback/src/cache.rs
+++ b/fluent-fallback/src/cache.rs
@@ -4,6 +4,7 @@ use std::{
     pin::Pin,
     task::Context,
     task::Poll,
+    task::Waker,
 };
 
 use crate::generator::{BundleIterator, BundleStream};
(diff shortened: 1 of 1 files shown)


emilio:
There's a bunch of unsafe in this file which could get some comments about why it's sound btw.

zbraniecki:
@djg - can you take a look at if this seems like the right fix for the issue I encountered in https://bugzilla.mozilla.org/show_bug.cgi?id=1613705#c56 ?

zbraniecki (approved):
r+ from Nika as well!

zbraniecki:
Thank you @emilio!

PR #225, opened 2021-05-13
Fix bug 1491867: Add ability to wrap pseudolocale strings in markers

mathjazz:


fluent-pseudo/src/lib.rs
```diff
@@ -23,7 +23,7 @@ static FLIPPED_CAPS_MAP: &[char] = &[
 static mut RE_EXCLUDED: Option<Regex> = None;
 static mut RE_AZ: Option<Regex> = None;
 
-pub fn transform_dom(s: &str, flipped: bool, elongate: bool) -> Cow<str> {
+pub fn transform_dom(s: &str, flipped: bool, elongate: bool, with_markers: bool) -> Cow<str> {
```
mathjazz, line 26: I was also thinking about "boundaries" instead of "markers".

fluent-pseudo/src/lib.rs
```diff
@@ -56,6 +56,11 @@ pub fn transform_dom(s: &str, flipped: bool, elongate: bool) -> Cow<str> {
     let result_range = pos + diff..result.len();
     let transform_sub = transform(&s[range], flipped, elongate);
     result.to_mut().replace_range(result_range, &transform_sub);
+
+    if with_markers {
+        return Cow::from("[") + result + Cow::from("]")
```
mathjazz, line 61: This is the first time I write Rust code. I assume there's more elegant way to do that?

mathjazz:
@dminor r?

No idea why the tests are failing...

fluent-pseudo/src/lib.rs
dminor:
```suggestion
        return Cow::from("[") + result + "]"
```
I had a look through the Cow documentation, and nothing really jumped out at me. @nordzilla Is there a nicer way of handling this?

dminor:
> @dminor r?
> 
> No idea why the tests are failing...

The tests pass locally for me with your changes (both stable and nightly). It is concerning that the crash is occurring in the fluent-pseudo test cases.

zbraniecki:
You don't have that many options and by the nature of this patch it is not critically performant do aiming for easy maintenance is the key, but if you wanted, two ways to "improve" the code could be:

* Switch API model to use Writer, this way you can populate it with whatever you want potentially without allocations
* Separate pre and post fix and feed chars into the result String.

The latter is probably what I would aim for but the difference is not important for this API. If you want you could look at how we handle PDI/FSI marks in fluent-bundle - https://github.com/projectfluent/fluent-rs/blob/master/fluent-bundle/src/resolver/pattern.rs#L62

I'd suggest adding more tests like - two literal parts separated by placeable, two literals side by side etc.

It would be also nice to consider if we want markers per string or per literal - @flodolo, @dminor , @mathjazz  - up to you!

flodolo:
> It would be also nice to consider if we want markers per string or per literal - @flodolo, @dminor , @mathjazz - up to you!

Sorry, could you give me an example of the two to understand?

zbraniecki:
> Sorry, could you give me an example of the two to understand?

Sure!
Example: `key = Hello, { $userName } to our product!`

Do we want `[Hello, John to our product!]` or maybe sometimes we may want to see `[Hello, ][John][ to our product]`?

Are message parts boundaries ever useful? Could they be to differentiate between `key = You have 5 messages` vs `key = You have { $count} messages`?

flodolo:
> Are message parts boundaries ever useful? Could they be to differentiate between `key = You have 5 messages` vs `key = You have { $count} messages`?

I think they risk being misleading? For example, would I be able to distinguish if `Hello, ][John][ to our product]` is the result of one string with a variable in the middle, or a collage of 3 strings?

On the other hand, it might be interesting to have something like `[Hello, {John} to our product]`, assuming that's possible.

dminor:
I think markers per string are fine for this PR.

Please do add some additional tests like Zibi suggested. You'll also need to bump the version number as part of this.

Since the tests only failed on Nightly, there's a chance the test failures were caused by the Nightly we were using when this was pushed. Unless we've pinned the Nightly version somewhere, in which case, we might want to try with a newer Nightly.

mathjazz:
Quick update: `transform()` is [called per `PatternElement`](https://searchfox.org/mozilla-central/source/third_party/rust/fluent-bundle/src/resolver/pattern.rs#34), so we can't really have per-Message markers easily.

Are we OK with that or should we refactor how `transform()` works?

Screenshot 2021-07-16 at 17 51 43

/cc @dminor @zbraniecki @flodolo

mathjazz:
Seems like the test is failing due to a bug in the coveralls version we use:
https://github.com/nickmerwin/node-coveralls/issues/280

Should I address it here or in a followup?

Other than that and the question raised above, this is ready for another look. Once the new version of fluent-pseudo is released, I can open a patch for the [Firefox implementation bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1711028).

dminor:
> Seems like the test is failing due to a bug in the coveralls version we use:
> [nickmerwin/node-coveralls#280](https://github.com/nickmerwin/node-coveralls/issues/280)
> 
> Should I address it here or in a followup?
> 
> Other than that and the question raised above, this is ready for another look. Once the new version of fluent-pseudo is released, I can open a patch for the [Firefox implementation bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1711028).

I think it would be best to keep the tests green, so please fix the coveralls version here as well.

dminor:
I'd also suggest we're fine with the current `transform()` behaviour, this is a step in the right direction, we can always iterate on it later.

mathjazz:
After I've updated the Changelog, coveralls no longer breaks. I believe my diagnosis from above is false - this is not related to the coveralls version. (We use the Coveralls GitHub Action from the Marketplace.)

Since the tests are green now, and since there are several similar test failures related to coveralls on master, I think we should address that problem separaretlly. @dminor Thoughts?

flodolo:
> That also means we can't have tests for literals separated by placeables- `transform_dom()` takes resolved strings as input (as opposed to fluent messages), so it can't really make a distinction between a literal and a placeable.

Double checking: does it mean that we can't have different markers for messages and placeables (variables, term references)?

The screenshot look good to me, it'd be great if I could visually differentiate between two potentially concatenated strings and a placeable in the middle of a message, but that's OK if that turns out to be too complicated.

mathjazz:
> Double checking: does it mean that we can't have different markers for messages and placeables (variables, term references)?

Correct.

The [Zibi's example](https://github.com/projectfluent/fluent-rs/pull/225#issuecomment-842554864) (`key = Hello, { $userName } to our product!`) will result in `[Hello, ][John][ to our product]`.

> The screenshot look good to me, it'd be great if I could visually differentiate between two potentially concatenated strings and a placeable in the middle of a message, but that's OK if that turns out to be too complicated.

That's good to know! We can still iterate on it later.

dminor:
> After I've updated the Changelog, coveralls no longer breaks. I believe my diagnosis from above is false - this is not related to the coveralls version. (We use the Coveralls GitHub Action from the Marketplace.)
> 
> Since the tests are green now, and since there are several similar test failures related to coveralls on master, I think we should address that problem separaretlly. @dminor Thoughts?

Makes sense to me!

dminor (approved):
LGTM, thanks!

PR #227, opened 2021-06-19
Replace ouroboros with self_cell

Voultapher:
Before (fluent-bundle):

$ cargo build
... 37 dependencies
Finished dev [unoptimized + debuginfo] target(s) in 18.39s

After:

$ cargo build
... 27 dependencies
Finished dev [unoptimized + debuginfo] target(s) in 14.59s

That's 26% faster in this very rough test. But the 10 fewer dependencies make this seem plausible. There are still some proc-macro dependencies like syn left, but they have nothing to do with `self_cell` as it has 0 dependencies and doesn't use proc-macros.

Disclaimer I'm the author of `self_cell`.

I ran `cargo test` and all non skipped tests passed before and after.

This is a continuation of https://github.com/projectfluent/fluent-rs/pull/221 which seems to have become stale.

I also ran the iai benchmarks:

Before:

```
iai_resolve_preferences
Instructions:             1875919 (No change)
L1 Accesses:              2669397 (No change)
L2 Accesses:                 9929 (No change)
RAM Accesses:                6519 (No change)
Estimated Cycles:         2947207 (No change)
```

After:

```
iai_resolve_preferences
Instructions:             1873556 (-0.125965%)
L1 Accesses:              2667079 (-0.086836%)
L2 Accesses:                 9848 (-0.815792%)
RAM Accesses:                6516 (-0.046019%)
Estimated Cycles:         2944379 (-0.095955%)
```

And it's slightly faster than ouroboros.

### DIFF
--- a/fluent-bundle/Cargo.toml
+++ b/fluent-bundle/Cargo.toml
@@ -29,11 +29,11 @@ include = [
 fluent-langneg = "0.13"
 fluent-syntax = { version = "0.11", path = "../fluent-syntax" }
 intl_pluralrules = "7.0.1"
-ouroboros = "0.9"
-smallvec = "1"
-unic-langid = "0.9"
 intl-memoizer = { version = "0.5", path = "../intl-memoizer" }
 rustc-hash = "1"
+self_cell = "0.9"
+smallvec = "1"
+unic-langid = "0.9"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


zbraniecki:
This looks great! Thank you so much for taking time to come up with a PR!

PR #231, opened 2021-07-28
Release fluent 0.16.0, which updates fluent-pseudo to 0.3

mathjazz:


mathjazz:
I filed #232 for the checks failure, which is unrelated to this patch.

zbraniecki:
Since fluent-pseudo has a breaking change, I think this should be 0.16.0.

PR #243, opened 2021-11-18
Allow optional resources

nordzilla:
- Add a new struct `ResourceId`.
    - `ResourceID`s can be optional or required.
- Update all code to use `ResourceId` instead of `String`.
- Add test scenario for empty resource in one locale.
- Add test scenario for empty resource in all locales.
- Add test scenario for missing optional resource in one locale.
- Add test scenario for missing optional resource in all locales.
- Add test scenario for missing required resource in one locale.
- Add test scenario for missing required resource in all locales.

### DIFF
--- a/fluent-resmgr/Cargo.toml
+++ b/fluent-resmgr/Cargo.toml
@@ -3,7 +3,7 @@ name = "fluent-resmgr"
 description = """
 Resource manager for Fluent localization resources.
 """
-version = "0.0.4"
+version = "0.0.5"
 authors = [
     "Zibi Braniecki <gandalf@mozilla.com>",
     "Staś Małolepszy <stas@mozilla.com>"
@@ -17,9 +17,9 @@ keywords = ["localization", "l10n", "i18n", "intl", "internationalization"]
 categories = ["localization", "internationalization"]
 
 [dependencies]
-fluent-bundle = "0.15"
+fluent-bundle = { path = "../fluent-bundle" }
+fluent-fallback = { path = "../fluent-fallback" }
 unic-langid = "0.9"
-fluent-fallback = { version = "0.5", path = "../fluent-fallback" }
 elsa = "1.3.2"
 futures = "0.3"
 
--- a/fluent-fallback/Cargo.toml
+++ b/fluent-fallback/Cargo.toml
@@ -4,7 +4,7 @@ description = """
 High-level abstraction model for managing localization resources
 and runtime localization lifecycle.
 """
-version = "0.5.0"
+version = "0.5.1"
 edition = "2018"
 authors = [
     "Zibi Braniecki <gandalf@mozilla.com>",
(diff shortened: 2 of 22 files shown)


zbraniecki (approved):
lgtm!

nordzilla:
@zbraniecki 

Thank you for taking a look and approving even though this was in draft.

I believe this is fully ready for review now. 

I do want to ask specifically about the changes I made to the `Cargo.toml` files. I largely changed these to refer to paths within this crate workspace, but some of them were previously pulling exact versions from crates.io.

Is there a reason for this? I feel like crates in this workspace should largely use the latest versions of each other unless there is a good reason not to. But I don't want to make this change lightly.

fluent-resmgr/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "fluent-resmgr"
 description = """
 Resource manager for Fluent localization resources.
 """
-version = "0.0.4"
+version = "0.0.5"
```
eemeli, line 6: Is there a reason not to use `0.1.0`? As is, there is no scope for differentiating major/minor/patch changes from each other.

zbraniecki (approved):
lgtm!

eemeli (commented):
I don't think I'm proficient enough in Rust yet to actually approve this, but it looks good to my amateur eyes.

fluent-resmgr/Cargo.toml
zbraniecki: Versions in Rust are meant to be conveying the subjective state of maturity of the package. 

`x.y` indicates production ready. `0.x.y` indicates `ready for use`, `0.0.x` indicates `POC`. `fluent-resmgr` is POC.

nordzilla:
@zbraniecki 

---

I have added an explainer section as you requested in https://github.com/mozilla/l10nregistry-rs/pull/19#pullrequestreview-825433903

---

I also explored the option of using `impl<S: Into<String>> From<(S, ResourceType)> for ResourceId` instead of a `ToResouceId` trait, and I'm fairly certain that I like the `ToResourceId` trait more for the following reasons:

These two blanket implementations cannot coexist:

```rust
impl<S: Into<String>> From<S> for ResouceId { ... }
impl<S: Into<String>> From<(S, ResourceType)> for ResourceId { ... }
```

Because someone else could implement
```rust
struct NewType {}
impl Fom<NewType> for String { ... }
impl From<(NewType, ResourceType)> for String { ... }
```

And this would create an ambiguous cycle with the above two implementations. 

--- 

Lastly,  regarding

> "One major bit that I'd like to clarify, is that you added a `ToResourceId` trait, but you don't generalize public method callers to be over `<R: ToResoureID>(res_ids: Vec<R>)`. Why?"

I did respond to this in the original comment, but I also want to respond here. Generalizing the public trait methods over a generic parameter would make the trait no longer object-safe, which I believe to be a big decision that I would prefer to make later only if needed. 

The current expecation that types are converted to `ResourceId` explicitly is forward-compatible if we decide to generalize over the trait at a later time, because of the blanket implementation. Since the blanket implementation `impl<T> From<T> for T` exists, then any old code that explicitly converts to `ResourceId` before passing it in will still work.

--- 

I'm getting approvals on the Gecko side of things, and I believe that I am ready to start publishing the crates and doing the vendoring.

zbraniecki (approved):
lgtm!

PR #251, opened 2022-02-28
Implement From<&String> for FluentValue

LegionMammal978:
This is useful for macros such as i18-embed-fl's [`fl!`](https://docs.rs/i18n-embed-fl/0.6.4/i18n_embed_fl/macro.fl.html) to accept `String` values by reference without requiring `as_str` or slice syntax. It directly corresponds to [`impl<'a> From<&'a String> for Cow<'a, str>`](https://doc.rust-lang.org/1.59.0/std/borrow/enum.Cow.html#impl-From%3C%26%27a%20String%3E).

### DIFF
--- a/fluent-bundle/src/types/mod.rs
+++ b/fluent-bundle/src/types/mod.rs
@@ -189,6 +189,12 @@ impl<'source> From<String> for FluentValue<'source> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


zbraniecki:
thank you! That looks good!

zbraniecki (approved):

PR #259, opened 2022-06-25
Add options for cardinal/ordinal plural types

oliver-ni:
Adds functionality for switching between plural types (cardinal and ordinal).

Fluent docs include this option in the `NUMBER` built-in function. Since `fluent-rs` doesn't have that yet, I included it in the place where all the other number options were. Please let me know if this isn't the appropriate place / way to include this functionality.

Also, I called it `kind` instead of `type` since the latter is a Rust keyword... a little weird but not sure if there's a better way to deal with that.

### DIFF
--- a/fluent-bundle/src/types/number.rs
+++ b/fluent-bundle/src/types/number.rs
@@ -9,6 +9,28 @@ use crate::args::FluentArgs;
 use crate::types::FluentValue;
 
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum FluentNumberKind {
+    Cardinal,
+    Ordinal,
+}
+
+impl std::default::Default for FluentNumberKind {
+    fn default() -> Self {
+        Self::Cardinal
+    }
+}
+
+impl From<&str> for FluentNumberKind {
+    fn from(input: &str) -> Self {
+        match input {
+            "cardinal" => Self::Cardinal,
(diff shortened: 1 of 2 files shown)


juliancoffee:
@oliver-ni to answer on using "type" in Rust, you can use raw identifiers.
https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html

fluent-bundle/src/types/number.rs
```diff
@@ -8,6 +8,28 @@ use intl_pluralrules::operands::PluralOperands;
 use crate::args::FluentArgs;
 use crate::types::FluentValue;
 
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum FluentNumberKind {
```
gregtatum, line 12: Suggestion: Please add docs here, with a link to the relevant Fluent documentation, and an example of what the `.ftl` syntax looks like.

gregtatum (changes requested):
The changes look reasonable, but I have a few requested changes.

> @oliver-ni to answer on using "type" in Rust, you can use raw identifiers.

Thought: I think I prefer `kind` here, it's used other places in the codebase.

Suggestion: This code is missing tests. It would probably be good to add some unit tests to the file, and see if there are relevant areas that an integration test could be added.

gregtatum:
Closing as stale.

PR #262, opened 2022-07-24
Resolve function instead in impl ResolveValue

### LINKED ISSUE #261: [Bug] Incorrect behaviour with following function invocation
Fluent Syntax book shows the following snippet.
```fluent
liked-count = { $num ->
        [0]     No likes yet.
        [one]   One person liked your message
       *[other] { $num } people liked your message
    }

liked-count2 = { NUMBER($num) ->
        [0]     No likes yet.
        [one]   One person liked your message
       *[other] { $num } people liked your message
    }
```
The quote is, "the following examples are equivalent and will both work.".
In reality, things are somewhat different.
Firstly, the first snippet depends on the type of argument passed as $num. If $num is a string, like "1", the result will be "1 people liked your message". If $num is number 1, the result is "One person liked your message". This is mostly ok.
But with the second snippet here is a real problem.
Considering we want to cast our number "1" to FluentValue::Number, we can call it to make the conversion. And it works on the playground.
https://projectfluent.org/play/?id=c9aca30ef4ba09384df54def5b5ead01

In fluent-bundle, it doesn't work. Even worse, it doesn't work even if we call NUMBER if $num is already a number (for example to add more options).
I created a gist that shows those 4 cases.
https://gist.github.com/juliancoffee/ab6f8bf691415f20942debfdfe50cc4b

Results are following:
1 people liked your message ("1")
One person liked your message (1)
1 people liked your message (NUMBER("1"))
1 people liked your message (NUMBER(1))

The problem I found is that we don't have functions properly in `impl ResolveValue` for `ast::InlineExpression`, and the fix is quite simple. I would gladly make a PR if that's appreciated.

juliancoffee:
Before this change, we were using the version from WriteValue, which
basically ignored the type of function and always converted it to string,
which made NUMBER function useless when used on strings or even wrong
when used on numbers.
Fixes #261

### DIFF
--- a/fluent-bundle/src/resolver/inline_expression.rs
+++ b/fluent-bundle/src/resolver/inline_expression.rs
@@ -171,6 +171,19 @@ impl<'p> ResolveValue for ast::InlineExpression<&'p str> {
                     FluentValue::Error
                 }
             }
+            Self::FunctionReference { id, arguments } => {
+                let (resolved_positional_args, resolved_named_args) =
+                    scope.get_arguments(Some(arguments));
+
+                let func = scope.bundle.get_entry_function(id.name);
+
+                if let Some(func) = func {
+                    let result = func(resolved_positional_args.as_slice(), &resolved_named_args);
+                    result
+                } else {
+                    FluentValue::Error
+                }
+            }
             _ => {
                 let mut result = String::new();
                 self.write(&mut result, scope).expect("Failed to write");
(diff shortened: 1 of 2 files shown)


juliancoffee:
@zbraniecki do you want to review this? :eyes:

juliancoffee:
@zbraniecki done.
I basically copied my example as a test by replacing `println!` with `assert_eq!`.
Let me know If I need to change something.

zbraniecki (approved):
thanks, lgtm!

Issue #260, opened 2022-07-20
[API Question] Why FluentBundle::format_pattern takes Option<'bundle FluentArgs>


juliancoffee:
format_pattern has the following signature:
 ```rust
pub fn format_pattern<'bundle>(
        &'bundle self,
        pattern: &'bundle ast::Pattern<&str>,
        args: Option<&'bundle FluentArgs>,
        errors: &mut Vec<FluentError>,
) -> Cow<'bundle, str>
```
And if I read lifetimes right, it means that args must have the same lifetime as a result. And it's impossible to create args at runtime because their lifetime will be obviously less than the bundle.
I have this code for i18n in my app.
```rust
pub struct Locale {
    langs: HashMap<LanguageIdentifier, FluentBundle<FluentResource>>,
    current_lang: LanguageIdentifier,
}

impl Locale {
    fn bundle(&self) -> &FluentBundle<FluentResource> {
        self.langs
            .get(&self.current_lang)
            .expect("failed to get bundle for current locale")
    }

    pub fn try_msg_ctx(
        &self,
        key: &str,
        args: &[(&str, &str)],
    ) -> Option<String> {
        let bundle = self.bundle();
        let msg = bundle.get_message(key)?;

        let context = wrap(args);

        let mut errs = Vec::new();
        let msg =
            bundle.format_pattern(msg.value()?, Some(&context), &mut errs);
        for err in errs {
            eprintln!("err: {err}")
        }
        
        // notice into_owned here
        Some(msg.into_owned())
    }
}
```
And because the result of try_msg_ctx borrows lifetime from args, I can't use Cow here and am forced to use String. This means I need to duplicate the code to get message with arguments and without arguments because in the latter case I don't need to create a new string, I can just get it from the bundle, but in the former case, I obviously need to create an owned string. And it seems that I should be ok with just using Cow<'bundle, string>, but because format_pattern links args and result, it doesn't compile (because args will always have a lesser lifetime).
So, my understanding is wrong or things actually can be improved so that args can have their own lifetime?

juliancoffee:
Btw, I think I solved my problem by reordering some lifetimes or just writing better code.
Instead of passing `&[(&str, &str)]` as args, I used FluentArgs directly and this function just works.
Here are bits of code.
```rust
pub struct Locale {
    langs: HashMap<LanguageIdentifier, FluentBundle<FluentResource>>,
    current_lang: LanguageIdentifier,
    fallback: LanguageIdentifier,
}

[...]

Issue #190, opened 2020-07-22
Arguments & ownership


vpzomtrrfrt:
Here's a simplified version of what I'm looking at, based on the example from the Good Practices page:
```rust
    let bundle: FluentBundle<FluentResource> = get_bundle_somehow();
    let mut errors = Vec::new();

    let num = get_number_somehow();

    let text = if num == 0 {
        let pattern = bundle.get_message("items-select").unwrap().value.unwrap();
        bundle.format_pattern(&pattern, None, &mut errors)
    } else {
        let pattern = bundle.get_message("items-selected").unwrap().value.unwrap();
        let args = std::iter::once(("num", num.into())).collect();
        bundle.format_pattern(&pattern, Some(&args), &mut errors)
    };

    println!("{}", text);
```

This doesn't compile, because the lifetime requirement on args is returned to the block:
```
error[E0597]: `args` does not live long enough
  --> src/main.rs:23:46
   |
17 |     let text = if num == 0 {
   |         ---- borrow later stored here
...
23 |         bundle.format_pattern(&pattern, Some(&args), &mut errors)
   |                                              ^^^^^ borrowed value does not live long enough
24 |     };
   |     - `args` dropped here while still borrowed
```

Is there something I'm missing here? How could this be made to work?

zbraniecki:
I think the issue here is that you want to return from the `else` a value that depends on the `args` while they're dropped at the end of it.

In #179 we're talking about switching the keys to be `String`, but the values of the args will still have lifetimes, so I'm not sure if you can workaround it.

zbraniecki:
@vpzomtrrfrt can you check if the current master solved it? We now accept `Cow` for `FluentArgs` which allows you to allocate there.

vpzomtrrfrt:
I don't think this is resolved, format_pattern still returns a value bound to the lifetime of the args

zbraniecki:
I see. The challenge here is that we're trying to be *really* light on how we handle allocations. That means that in theory, your `items-selected` may look like this:

```
items-selected = { $userName }
```

in which case we want to take `userName` which will be ` Cow<str>` as `FluentValue::String` *and return it* untouched. For that to happen, we need the same lifetime on `userName` and on return value from `format_pattern`.

For numbers, we perform formatting so in theory maybe we could tailor the lifetimes to not span to other `FluentValue` variants, but that's complicated.

If you are ok with owning the result, you can just do:

```rust
    let text = if num == 0 {
[...]

PR #264, opened 2022-07-27
Fix FluentBundle::format_pattern lifetimes

MathieuTricoire:
Hi! First of all, thank you for all the work put into these crates!

I have faced the issue where a formatted pattern cannot outlive its arguments (same issue raised by #190 and #260)

It would be useful to allow it, since the returned `Cow` doesn't seems to be tied to any arguments references.

_Aside note_
It was the case before at the time of the issue #190 i.e. https://github.com/projectfluent/fluent-rs/blob/43f1f59534487ca539164a160498138a6ae641b2/fluent-bundle/src/resolve.rs#L146-L148 but current code only return reference to the resource when there is only 1 `TextElement` i.e. https://github.com/projectfluent/fluent-rs/blob/8e17ee622bb98aad2ff2382755633148637fede9/fluent-bundle/src/resolver/pattern.rs#L94-L101
_End of the aside note_

The problem is this code doesn't compile:
```rust
use fluent_bundle::{FluentArgs, FluentBundle, FluentResource};
use std::borrow::Cow;
use unic_langid::langid;

fn main() {
    let res = FluentResource::try_new("greeting = Hello, { $name }".to_string()).unwrap();
    let en_us = langid!("en-US");
    let mut bundle = FluentBundle::new(vec![en_us]);
    bundle.add_resource(res).unwrap();

    let result = greeting("World!", &bundle);
    println!("{result}");
}

fn greeting<'a, 'b>(name: &'a str, bundle: &'b FluentBundle<FluentResource>) -> Cow<'b, str> {
    let mut errors = vec![];
    let value = bundle.get_message("greeting").unwrap().value().unwrap();

    args.set("name", name);

    bundle.format_pattern(value, Some(&args), &mut errors)
}
```

Error:
```
error[E0623]: lifetime mismatch
  --> src/main.rs:22:5
   |
15 | fn greeting<'a, 'b>(name: &'a str, bundle: &'b FluentBundle<FluentResource>) -> Cow<'b, str> {
   |                           -------                                               ------------
   |                           |
   |                           this parameter and the return type are declared with different lifetimes...
...
22 |     bundle.format_pattern(value, Some(&args), &mut errors)
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ...but data from `name` is returned here
```

So the only issues is about lifetimes and this PR is a proposition to fix that.

Let me know if I'm correct, if you agree with the lifetimes naming and the tests.

Thank again!

---

FYI I tried to differentiate more lifetimes in `Scope` i.e. `&'bundle FluentBundle` and `ast::Pattern<&'source str>` because theorically they are different lifetimes but it requires too much work because we don't have access to `'source` lifetime from `FluentResource` and it's most of time obfuscated as `R` when referred to.
It could have been interesting because the `format_pattern` should return `Cow<'source, str>` and could theorically outlives `FluentBundle` but not the `FluentResource`.

```rust
pub fn format_pattern<'bundle, 'source, 'args>(
   &'bundle self,
   pattern: &'bundle ast::Pattern<&'source str>,
   args: Option<&'args FluentArgs>,
   errors: &mut Vec<FluentError>,
) -> Cow<'source, str>;
```

### DIFF
--- a/fluent-bundle/src/types/mod.rs
+++ b/fluent-bundle/src/types/mod.rs
@@ -186,10 +186,7 @@ impl<'source> FluentValue<'source> {
         match self {
             FluentValue::String(str) => FluentValue::String(Cow::from(str.to_string())),
             FluentValue::Number(s) => FluentValue::Number(s.clone()),
-            FluentValue::Custom(s) => {
-                let new_value: Box<dyn FluentType + Send> = s.duplicate();
-                FluentValue::Custom(new_value)
-            }
+            FluentValue::Custom(s) => FluentValue::Custom(s.duplicate()),
             FluentValue::Error => FluentValue::Error,
             FluentValue::None => FluentValue::None,
         }


juliancoffee:
(not a maintainer here, but I was able to fix my code by passing args outside, your code can be fixed like that)
```rust
use fluent_bundle::{FluentArgs, FluentBundle, FluentResource};
use std::borrow::Cow;
use unic_langid::langid;

fn main() {
    let res = FluentResource::try_new("greeting = Hello, { $name }".to_string()).unwrap();
    let en_us = langid!("en-US");
    let mut bundle = FluentBundle::new(vec![en_us]);
    bundle.add_resource(res).unwrap();

    args.set("name", "World!");
    let result = greeting(&args, &bundle);

    println!("{result}");
}

fn greeting<'a>(args: &'a FluentArgs, bundle: &'a FluentBundle<FluentResource>) -> Cow<'a, str> {
    let mut errors = vec![];
    let value = bundle.get_message("greeting").unwrap().value().unwrap();

    bundle.format_pattern(value, Some(args), &mut errors)
}
```

MathieuTricoire:
Yes I know, sorry if my example is trivial, I gave a simple example to illustrate the problem and of course it can be solved as you suggest but my use case is more complex and cannot be solved this way.

My point is that there is actually no reason for this limitation, the returned `Cow` will never contain a reference to an argument (if my understanding of the code is correct).

The only time the returned `Cow` will contain a reference to a `str` will be if the given pattern contains only one element and that element is an `ast::PatternElement::TextElement`, then it will directly return the reference to the value of the ast element which is a great thing for performance reasons to not allocate a new `String`
But in other situations a `String` will be allocated and the given pattern will be interpolated and returned, so the `Cow` will contain the allocated `String`.

So no reason to tie the returned lifetime to the arguments lifetime, a fix that will open up some interesting new possibilities IMHO.

The code is actually very easy to read to understand the issue:
https://github.com/projectfluent/fluent-rs/blob/df81ab8bd772dd5a54f2b84378b2a33197e94ace/fluent-bundle/src/resolver/pattern.rs#L83-L107

MathieuTricoire:
Hi @zbraniecki, would it be possible for someone to have a look at this PR and also #271?

I need this for a crate I have published: https://github.com/MathieuTricoire/l10n

Don't hesitate to let me know your thoughts on these 2 PRs and also on my crate if you want and have time.

Thanks again for all your work!

zbraniecki:
Hi, I'm sorry for the delay. I haven't had a lot of time to devote to fluent-rs lately due to ICU4X 1.0 release.

@eemeli @nordzilla @gregtatum @dminor - would any of you have a moment to look at this?

if not, I'll try to get to it in a couple weeks.

fluent-bundle/src/types/mod.rs
```diff
@@ -181,6 +181,19 @@ impl<'source> FluentValue<'source> {
             FluentValue::None => "".into(),
         }
     }
+
+    pub fn into_owned<'a>(&self) -> FluentValue<'a> {
+        match self {
+            FluentValue::String(str) => FluentValue::String(Cow::from(str.to_string())),
+            FluentValue::Number(s) => FluentValue::Number(s.clone()),
+            FluentValue::Custom(s) => {
+                let new_value: Box<dyn FluentType + Send> = s.duplicate();
```
gregtatum, line 190: Question: Does the compiler want you to put this on the stack as `new_value` and provide a type or could you turn this into a one-line? It's unclear from just reading the code if there is a reason for the extra verbosity.

`FluentValue::Custom(s.duplicate())`

fluent-bundle/src/resolver/scope.rs
```diff
@@ -8,28 +8,28 @@ use std::borrow::Borrow;
 use std::fmt;
 
 /// State for a single `ResolveValue::to_value` call.
-pub struct Scope<'scope, 'errors, R, M> {
+pub struct Scope<'bundle, 'ast, 'args, 'errors, R, M> {
```
gregtatum, line 11: Thought: This is a little gross with the proliferation of the lifetimes, but it seems like a reasonable approach to solve the issue. It at least makes it more explicit.

gregtatum: Thought: Reading through it a few times, it at least makes things very explicit for each reference.

fluent-bundle/src/resolver/inline_expression.rs
```diff
@@ -147,11 +147,11 @@ impl<'p> WriteValue for ast::InlineExpression<&'p str> {
...
-impl<'p> ResolveValue for ast::InlineExpression<&'p str> {
-    fn resolve<'source, 'errors, R, M>(
-        &'source self,
-        scope: &mut Scope<'source, 'errors, R, M>,
-    ) -> FluentValue<'source>
+impl<'bundle> ResolveValue<'bundle> for ast::InlineExpression<&'bundle str> {
+    fn resolve<'ast, 'args, 'errors, R, M>(
+        &'ast self,
+        scope: &mut Scope<'bundle, 'ast, 'args, 'errors, R, M>,
+    ) -> FluentValue<'bundle>
```
gregtatum, line 154: Praise: `'bundle` is much clearer than `'p` on why a lifetime exists, and how the reference is constrained.

fluent-bundle/src/resolver/inline_expression.rs
```diff
@@ -160,16 +160,18 @@ impl<'p> ResolveValue for ast::InlineExpression<&'p str> {
             Self::StringLiteral { value } => unescape_unicode_to_string(value).into(),
             Self::NumberLiteral { value } => FluentValue::try_number(*value),
             Self::VariableReference { id } => {
-                let args = scope.local_args.as_ref().or(scope.args);
```
gregtatum, line 163: Thought: These changes seem reasonable reading through it.

Question: Do you know if this code is explicitly covered by tests? I'd like to doubly make sure there is no behavior change here.

fluent-bundle/tests/bundle.rs
```diff
@@ -19,19 +18,75 @@ fn add_resource_override() {
...
 
     bundle.add_resource_overriding(&res2);
 
     let value = bundle
         .get_message("key")
-        .expect("Failed to retireve a message")
+        .expect("Failed to retrieve a message")
         .value()
-        .expect("Failed to retireve a value of a message");
+        .expect("Failed to retrieve a value of a message");
```
gregtatum, line 32: praise: Thanks for the drive-by fixes.

fluent-bundle/tests/bundle.rs
```diff
@@ -19,19 +18,75 @@ fn add_resource_override() {
...
+fn arguments_outlive_formatted_pattern() {
+    let res = FluentResource::try_new("key = { $variable }".to_string()).unwrap();
+    let en_us = langid!("en-US");
+
+    let mut bundle = FluentBundle::new(vec![en_us]);
+    bundle.add_resource(&res).expect("Failed to add a resource");
+
+    let mut errors = vec![];
+
+    let formatted_pattern = {
```
gregtatum, line 78: Thought: Yes this makes for a very convincing example of why this change is needed.

gregtatum (approved):
Thanks for changes, it looks like a great addition. I left one question I'd like answered before merge, but I'm happy to approve it. Also, please note that my review is using [conventional comments](https://conventionalcomments.org/) which uses labels for which comments are actionable. I believe I didn't include any `suggestion:` or `nitpick:` comments, so things look good.

fluent-bundle/src/resolver/inline_expression.rs
MathieuTricoire: I've done some experimentation right now to see if this is properly covered (deleting and swapping branches in the if condition) and there are tests that cover this in "fixtures".

I vaguely remember that I have look at related tests when touching this code to have a better grasp, if I'm not wrong the "Parameterized term attributes" test suite in `tests/fixtures/macros.yaml` are related to this, for example swapping if branches break this test: https://github.com/projectfluent/fluent-rs/blob/c91d9f48b5a8734bd9183145320f5108526b2d1e/fluent-bundle/tests/fixtures/macros.yaml#L358-L364

Do you think it's ok like this?

MathieuTricoire:
Thanks @gregtatum for all the feedback on my PRs 👍

fluent-bundle/src/resolver/inline_expression.rs
gregtatum: Good enough for merging I think!

gregtatum:
> Thanks @gregtatum for all the feedback on my PRs 👍

Thanks for the contributions!

MathieuTricoire:
I didn't want to open an issue for that so I'm writing here. Just wanted to know if you have a timeline to release fluent-bundle with the latest PRs?

PR #273, opened 2022-10-13
Allow optional arguments

MathieuTricoire:
Any thoughts about implementing `From` for `Option`?:

```rust
impl<T> From<Option<T>> for FluentValue
where
    T: Into<FluentValue>
```

I find interesting to allow something like this:
```fluent
busy = { $gender ->
    [male] Occupé
    [female] Occupée
   *[other] Non disponible
}
```

```rust
let gender: Option<String> = None;
args.set("gender", gender.into());
let value = bundle.format_pattern(&pattern, Some(&args), &mut errors);
```

### DIFF
--- a/fluent-bundle/src/types/mod.rs
+++ b/fluent-bundle/src/types/mod.rs
@@ -206,3 +206,15 @@ impl<'source> From<Cow<'source, str>> for FluentValue<'source> {
         FluentValue::String(s)
     }
 }
+
+impl<'source, T> From<Option<T>> for FluentValue<'source>
+where
+    T: Into<FluentValue<'source>>,
+{
+    fn from(s: Option<T>) -> Self {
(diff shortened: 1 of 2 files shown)


gregtatum (changes requested):
It seems reasonable to me. I'm marking request changes for a simple unit test. Most the existing tests are more on the integration level, but a unit test here should be good enough for merging.

MathieuTricoire:
> It seems reasonable to me. I'm marking request changes for a simple unit test. Most the existing tests are more on the integration level, but a unit test here should be good enough for merging.

I've added an integration test, I think it fits more with what exists, is that ok?

gregtatum (approved):
Thanks for tests, that makes it very clear what's going on here, and thanks for the contribution!

PR #274, opened 2022-10-27
Add more documentation to fluent-bundle

gregtatum:
I wrote more docs as a way to learn the `fluent-bundle` codebase.

### DIFF
--- a/fluent-bundle/src/args.rs
+++ b/fluent-bundle/src/args.rs
@@ -3,9 +3,17 @@ use std::iter::FromIterator;
 
 use crate::types::FluentValue;
 
-/// A map of arguments passed from the code to
-/// the localization to be used for message
-/// formatting.
+/// Fluent messages can use arguments in order to programmatically add values to a
+/// translated string. For instance, in a localized application you may wish to display
+/// a user's email count. This could be done with the following message.
+///
+/// `msg-key = Hello, { $user }. You have { $emailCount } messages.`
+///
+/// Here `$user` and `$emailCount` are the arguments, which can be filled with values.
+///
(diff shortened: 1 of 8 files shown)


fluent-bundle/src/args.rs
```diff
@@ -48,14 +56,17 @@ use crate::types::FluentValue;
 pub struct FluentArgs<'args>(Vec<(Cow<'args, str>, FluentValue<'args>)>);
 
 impl<'args> FluentArgs<'args> {
+    /// Creates a new empty list.
```
dminor, line 59:
```suggestion
    /// Creates a new empty argument map.
```

fluent-bundle/src/resolver/mod.rs
```diff
@@ -25,7 +31,9 @@ pub(crate) trait ResolveValue {
         M: MemoizerKind;
 }
 
+/// Resolves and AST node to a string that is written to source `W`.
```
dminor, line 34:
```suggestion
/// Resolves an AST node to a string that is written to source `W`.
```

fluent-bundle/src/resolver/mod.rs
```diff
@@ -25,7 +31,9 @@ pub(crate) trait ResolveValue {
...
 pub(crate) trait WriteValue {
+    /// Resolves and AST node to a string that is written to source `W`.
```
dminor, line 36:
```suggestion
    /// Resolves an AST node to a string that is written to source `W`.
```

gregtatum:
Rebasing due to conflicts.

PR #277, opened 2022-11-04
Change the resource ids to be a Set rather than a Vec

gregtatum:
See [Bug 1755216](https://bugzilla.mozilla.org/show_bug.cgi?id=1755216) for a larger discussion, but this changes the behavior so that when multiple `ResourceId`s are added, they do not trigger a Fluent error. It changes the behavior to a noop, since the behavior of the translated messages doesn't actually change.

### DIFF
--- a/fluent-fallback/src/localization.rs
+++ b/fluent-fallback/src/localization.rs
@@ -5,6 +5,7 @@ use crate::{
     types::ResourceId,
 };
 use once_cell::sync::OnceCell;
+use rustc_hash::FxHashSet;
 use std::rc::Rc;
 
 pub struct Localization<G, P>
@@ -16,7 +17,7 @@ where
     generator: G,
     provider: P,
     sync: bool,
-    res_ids: Vec<ResourceId>,
+    res_ids: FxHashSet<ResourceId>,
 }
 
 impl<G, P> Localization<G, P>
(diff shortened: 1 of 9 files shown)


fluent-fallback/src/localization.rs
```diff
@@ -16,21 +17,24 @@ where
...
-    res_ids: Vec<ResourceId>,
+    res_ids: FxHashSet<ResourceId>,
 }
 
 impl<G, P> Localization<G, P>
 where
     G: BundleGenerator<LocalesIter = P::Iter> + Default,
     P: LocalesProvider + Default,
 {
-    pub fn new(res_ids: Vec<ResourceId>, sync: bool) -> Self {
```
gregtatum, line 27: I made this method generic rather than requiring the `FxHashSet` to be part of the public API.

fluent-fallback/tests/localization_test.rs
```diff
@@ -488,3 +497,32 @@ async fn localization_handle_state_changes_mid_async() {
...
+        .format_value_sync("missing-message", None, &mut errors)
+        .unwrap();
+    assert_eq!(value, None);
+
+    let value = bundles
+        .format_value_sync("hello-world-3", None, &mut errors)
+        .unwrap();
+    assert_eq!(value, Some(Cow::Borrowed("Hello World 3 [en]")));
+
+    assert_eq!(errors.len(), 4);
```
dminor, line 527: Suggestion: Can we assert on the contents of the errors here? I'm not clear on what errors we are expecting. From the description in Bug 1755216 it sounds like we want to suppress any errors from adding `test2.ftl` a second time, but allow other errors through.

dminor (approved):
This makes sense to me in the context of the discussion in Bug 1755216.

fluent-fallback/tests/localization_test.rs
gregtatum: I was lazy and copied and pasted a pre-existing test. I simplified it so that there are no errors.

Issue #278, opened 2022-11-08
ResourceManager::get_resource panics rather than returning a result


gregtatum:
This API should let the consumer decide what to do when a resource can't be retrieved. Currently it panics with an `unwrap`.

https://github.com/projectfluent/fluent-rs/blob/9e62af6f18f5efed4ad8faaca2c6c0ceb14d5bde/fluent-resmgr/src/resource_manager.rs#L38

adriantombu:
I can have a look at that 👀

gregtatum:
Thanks! I'll assign it to you.

gregtatum:
Oh, and just be aware I'm documenting around this area, so there might be some short-term merge conflicts.

PR #282, opened 2022-11-09
Return a result for ResourceManager::get_resource

adriantombu:
See #278 for more context

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,6 @@ members = [
 exclude = [
     "fluent-cli",
 ]
+
+[workspace.dependencies]
+thiserror = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,3 @@ members = [
 exclude = [
     "fluent-cli",
 ]
-
-[workspace.dependencies]
-thiserror = "1.0"
--- a/fluent-resmgr/Cargo.toml
+++ b/fluent-resmgr/Cargo.toml
@@ -23,7 +23,7 @@ unic-langid = "0.9"
 elsa = "1.5"
 futures = "0.3"
 rustc-hash = "1"
-thiserror = "1.0.37"
+thiserror.workspace = true
 
 [dev-dependencies]
 unic-langid = { version = "0.9", features = ["macros"]}
--- a/fluent-resmgr/Cargo.toml
+++ b/fluent-resmgr/Cargo.toml
@@ -23,7 +23,7 @@ unic-langid = "0.9"
 elsa = "1.5"
 futures = "0.3"
 rustc-hash = "1"
-thiserror.workspace = true
+thiserror = "1.0"
 
 [dev-dependencies]
 unic-langid = { version = "0.9", features = ["macros"]}
--- a/fluent-resmgr/Cargo.toml
+++ b/fluent-resmgr/Cargo.toml
@@ -23,6 +23,7 @@ unic-langid = "0.9"
 elsa = "1.5"
 futures = "0.3"
 rustc-hash = "1"
+thiserror = "1.0.37"
 
 [dev-dependencies]
 unic-langid = { version = "0.9", features = ["macros"]}
--- a/fluent-resmgr/src/resource_manager.rs
+++ b/fluent-resmgr/src/resource_manager.rs
@@ -6,8 +6,8 @@ use fluent_fallback::{
 };
 use futures::stream::Stream;
 use rustc_hash::FxHashSet;
-use std::fs;
 use std::io;
+use std::{fs, iter};
 use thiserror::Error;
 use unic_langid::LanguageIdentifier;
 
@@ -57,15 +57,14 @@ impl ResourceManager {
             .path_scheme
             .replace("{locale}", locale)
             .replace("{res_id}", res_id);
-        Ok(if let Some(res) = self.resources.get(&path) {
-            res
+        Ok(if let Some(resource) = self.resources.get(&path) {
+            resource
         } else {
-            let string = read_file(&path)?;
-            let res = match FluentResource::try_new(string) {
-                Ok(res) => res,
-                Err((res, _err)) => res,
+            let resource = match FluentResource::try_new(read_file(&path)?) {
+                Ok(resource) => resource,
+                Err((resource, _err)) => resource,
             };
-            self.resources.insert(path.to_string(), Box::new(res))
+            self.resources.insert(path.to_string(), Box::new(resource))
         })
     }
 
@@ -81,20 +80,21 @@ impl ResourceManager {
     ) -> Result<FluentBundle<&FluentResource>, Vec<ResourceManagerError>> {
         let mut errors: Vec<ResourceManagerError> = vec![];
         let mut bundle = FluentBundle::new(locales.clone());
+        let locale = &locales[0];
 
-        resource_ids.iter().for_each(|res_id| {
-            let _ = self
-                .get_resource(res_id, &locales[0].to_string())
-                .map_err(|err| errors.push(err))
-                .and_then(|res| {
-                    bundle.add_resource(res).map_err(|errs| {
+        for resource_id in &resource_ids {
+            match self.get_resource(resource_id, &locale.to_string()) {
+                Ok(resource) => {
+                    if let Err(errs) = bundle.add_resource(resource) {
                         errs.into_iter()
-                            .for_each(|err| errors.push(ResourceManagerError::Fluent(err)))
-                    })
-                });
-        });
+                            .for_each(|error| errors.push(ResourceManagerError::Fluent(error)))
+                    }
+                }
+                Err(error) => errors.push(error),
+            };
+        }
 
-        if errors.iter().len() > 0 {
+        if !errors.is_empty() {
             Err(errors)
         } else {
             Ok(bundle)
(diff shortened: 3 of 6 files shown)


adriantombu:
@gregtatum just creating a draft here to get early feedback and see if I'm going in the right direction ^^

fluent-resmgr/Cargo.toml
```diff
@@ -22,6 +22,7 @@ fluent-fallback = { version = "0.6.0", path = "../fluent-fallback" }
 unic-langid = "0.9"
 elsa = "1.5"
 futures = "0.3"
+thiserror = "1.0.37"
```
gregtatum, line 25: Suggestion: The version number should match the other `thiserror` version.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -27,58 +27,90 @@ impl ResourceManager {
...
+        if errors.iter().len() > 0 {
+            Err(errors)
+        } else {
+            Ok(res.into_iter())
+        }
     }
 }
 
+#[derive(Error, Debug)]
+pub enum ResourceManagerError {
```
gregtatum, line 106: Suggestion: This will need some documentation since it's public.

fluent-resmgr/tests/localization_test.rs
```diff
@@ -36,11 +36,13 @@ fn localization_format_value() {
...
 
-    let bundle = res_mgr.get_bundle(vec![langid!("en-US")], vec!["test.ftl".into()]);
+    let bundle = res_mgr
+        .get_bundle(vec![langid!("en-US")], vec!["test.ftl".into()])
+        .expect("Could not get bundle");
 
     let mut errors = vec![];
     let msg = bundle.get_message("hello-world").expect("Message exists");
     let pattern = msg.value().expect("Message has a value");
-    let value = bundle.format_pattern(&pattern, None, &mut errors);
```
gregtatum, line 44: Question: These `&` changes seem unrelated drive-by fixes. I'm assuming since the `pattern` here is already a reference.

gregtatum (commented):
Yeah this seems to be in the right direction. I did a quick read-through with a few comments. It will need tests. I have some I wrote on some local patches working on some code coverage (that's where this issue came from) but feel free to add your own, or wait a day or two until I get mine online to base it off of.

fluent-resmgr/tests/localization_test.rs
adriantombu: Yeah, I did a few `clippy` quick fixes, I thought it wouldn't hurt much 😄

fluent-resmgr/Cargo.toml
adriantombu: Is this ok if we start using workspace dependencies to simplify the versions across the crates? See this commit https://github.com/projectfluent/fluent-rs/pull/282/commits/a75bf0858e21b88a61292122b14efb34caae07ba

fluent-resmgr/src/resource_manager.rs
adriantombu: Added here https://github.com/projectfluent/fluent-rs/pull/282/commits/5ad427088f39215d19f2892c824f6155598677ef

adriantombu:
@gregtatum I just rebased and fixed the tests you wrote to make them work with my changes. Do you see anything else to do on this?

fluent-resmgr/Cargo.toml
gregtatum: This might be better as a separate PR. I don't know if the [Firefox vendoring process](https://firefox-source-docs.mozilla.org/build/buildsystem/rust.html) works with workspace, and we would need to block on verifying that it would work. I wouldn't want to break the Firefox integration.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -48,21 +48,25 @@ impl ResourceManager {
...
+        &self,
+        res_id: &str,
+        locale: &str,
+    ) -> Result<&FluentResource, ResourceManagerError> {
         let path = self
             .path_scheme
             .replace("{locale}", locale)
-            .replace("{res_id}", resource_id);
-        if let Some(res) = self.resources.get(&path) {
+            .replace("{res_id}", res_id);
```
gregtatum, line 59: Nit: I'd prefer to have the full name written out, as `res` can be a little mysterious. I know previous authors used smaller names, but I find being explicit makes the code so much easier to read and understand when you have low context for changes.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -74,14 +78,27 @@ impl ResourceManager {
...
-    ) -> FluentBundle<&FluentResource> {
+    ) -> Result<FluentBundle<&FluentResource>, Vec<ResourceManagerError>> {
+        let mut errors: Vec<ResourceManagerError> = vec![];
         let mut bundle = FluentBundle::new(locales.clone());
-        for res_id in &resource_ids {
-            println!("res_id {:?}", res_id);
-            let res = self.get_resource(res_id, &locales[0].to_string());
-            bundle.add_resource(res).unwrap();
+
+        resource_ids.iter().for_each(|res_id| {
```
gregtatum, line 85: Nit: I don't think anywhere else has used the `for_each`. I think it would be simpler and more idiomatic to use a `for` loop here.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -74,14 +78,27 @@ impl ResourceManager {
...
+            let _ = self
```
gregtatum, line 86: Nit: `let _` isn't needed here.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -74,14 +78,27 @@ impl ResourceManager {
...
+                .map_err(|err| errors.push(err))
+                .and_then(|res| {
+                    bundle.add_resource(res).map_err(|errs| {
+                        errs.into_iter()
+                            .for_each(|err| errors.push(ResourceManagerError::Fluent(err)))
+                    })
+                });
+        });
+
+        if errors.iter().len() > 0 {
```
gregtatum, line 97: Nit: This could be simplified to `errors.is_empty()`.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -92,24 +109,41 @@ impl ResourceManager {
...
-                for resource_id in &resource_ids {
-                    let resource = res_mgr.get_resource(resource_id, &locale.to_string());
-                    bundle.add_resource(resource).unwrap();
-                }
-                bundle
-            })
-        })
+    ) -> Result<impl Iterator<Item = FluentBundle<&FluentResource>>, Vec<ResourceManagerError>>
+    {
+        let mut ptr = 0;
```
gregtatum, line 114: Nit: `idx` is clearer choice here compared to `ptr`. I think this may have been missed in a rebase.

fluent-resmgr/src/resource_manager.rs
```diff
@@ -92,24 +109,41 @@ impl ResourceManager {
...
+        let mut errors: Vec<ResourceManagerError> = vec![];
+        let mut res = vec![];
+
+        while let Some(locale) = locales.get(ptr) {
```
gregtatum, line 118: Issue: This changes the behavior of this function from being lazy (where the bundles are loaded on each iteration),  to fully evaluating the bundles. The intent for this iterator is to only evaluate the bundle on iteration.

Rather than returning ` Result<Iterator>` this should return a:

`impl Iterator<Item = Result< ... >>`

fluent-resmgr/src/resource_manager.rs
gregtatum: Praise: Thanks for the docs!

fluent-syntax/Cargo.toml
```diff
@@ -27,7 +27,7 @@ include = [
 [dependencies]
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { version = "1.0", optional = true }
-thiserror = "1.0"
+thiserror.workspace = true
```
gregtatum, line 30: Suggestion: See above on the workspace feedback.

gregtatum (changes requested):
Thanks for the updates on the patch! This is real close!

I'm marking request changes for a few suggestions and nitpicks that I had.

I also found one issue, where the `get_bundles` method had a change in behavior.

Once you are ready for a re-review please click the little "refresh" icon next to my name in the review, as this will re-trigger review for me. I missed the comments that you had left in my notifications deluge. Setting the review status will make sure that it's in my queue to take a look at.

Cargo.toml
```diff
@@ -13,3 +13,6 @@ members = [
 exclude = [
     "fluent-cli",
 ]
+
+[workspace.dependencies]
+thiserror = "1.0"
```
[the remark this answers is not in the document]
adriantombu: Makes sense, I'll provide another PR for this so that we are not blocked by it!

fluent-resmgr/src/resource_manager.rs
adriantombu: Clippy was unhappy about it, but I switched to a `match` to avoid this ugly `let _`

fluent-resmgr/src/resource_manager.rs
adriantombu: Wooops, sorry about that 🙈

gregtatum (approved):
Thanks for the contribution!

PR #284, opened 2022-11-09
Write up the release process

gregtatum:


eemeli:
I'm investigating how to add them to the org so that becomes possible.

docs/release-process.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+Upgrading fluent-fallback from 0.6.0 to 0.7.0
+ Updating fluent-resmgr's dependency from 0.6.0 to 0.7.0
+ Updating fluent-testing's dependency from 0.6.0 to 0.7.0
+```
+
+After this the depending packages must be updated as well:
+
+```.sh
```
nordzilla, line 25: > Question

Is this supposed to be `.sh` instead of just `sh`?

nordzilla (approved):
Looks good to me, just one formatting question.

PR #289, opened 2022-11-14
Add a github workflow to run rust fmt

### LINKED ISSUE #288: rustmt should be run in CI
Rustfmt should be added to the github workflows in `.github/workflows`

gybrish:
Resolves #288 

Tested locally with act:

```
[Rustfmt/format]   🐳  docker exec cmd=[node /var/run/act/actions/mbrobbel-rustfmt-check@master/dist/index.js] user= workdir=
| [command]/root/.cargo/bin/cargo + fmt -- -l
[Rustfmt/format]   ✅  Success - Main mbrobbel/rustfmt-check@master
[Rustfmt/format] 🏁  Job succeeded
```

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -3,7 +3,7 @@ name: Rustfmt
 on:
   # Triggers the workflow on push request events but only for the master branch
   push:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   format:


.github/workflows/format.yml
```diff
@@ -0,0 +1,20 @@
...
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            components: rustfmt
+            override: true
+      - uses: mbrobbel/rustfmt-check@master
```
gregtatum, line 18: Suggestion: I'd prefer not to rely on a small third-party CI job, as these can be sources of breakage and security issues. In ICU4X we built out the CI to to run the command manually. There is a bit of misdirection in the following links around the use of cargo make which I think you could ignore and put the format command directly in the yml file, but rely on the rustup commands.

https://github.com/unicode-org/icu4x/blob/d418731826aea10cbc72d6a3b73789a981e425bd/.github/workflows/build-test.yml#L513-L550

https://github.com/unicode-org/icu4x/blob/d418731826aea10cbc72d6a3b73789a981e425bd/tools/scripts/tidy.toml#L14-L18

gregtatum (changes requested):
Thanks for the PR! I have some feedback around using a smaller repo for github actions.

.github/workflows/format.yml
gybrish: I've removed the dependency on `mbrobbel/rustfmt-check`. There might be an argument about including the fmt check step with the `test.yml` actions as it also needs the rust toolchain download/install but imo it's cleaner if it's a separate job.

.github/workflows/format.yml
```diff
@@ -0,0 +1,21 @@
+name: Rustfmt
+
+on:
+  # Triggers the workflow on push request events but only for the master branch
+  push:
+    branches: [ master ]
```
gregtatum, line 6: Issue: I renamed the `master` branch to `main` yesterday on #290.

gregtatum (changes requested):
Praise: Thanks for handling my review feedback! The solution looks good.

I'd like to manually test this, and I believe the `master` to `main` change needs to happen before I do that, so I'm marking "request changes" so that the status is clear. Once you do that I can run the CI and merge it in.

gybrish:
I think we should be good now!

I had some trivial dry run tests locally, just messed with the formatting to make sure the check was working correctly:
```
|      type Item = &'a I::Item;
|      type IntoIter = CacheIter<'a, I, R>;
|
| -    fn into_iter(self) -> Self::IntoIter {CacheIter {
| +    fn into_iter(self) -> Self::IntoIter {
| +        CacheIter {
|              cache: self,
|              curr: 0,
|          }
[Rustfmt/format]   ❌  Failure - Main Run rustfmt
[Rustfmt/format] exitcode '1': failure
[Rustfmt/format] 🏁  Job failed
```

gregtatum (approved):

gregtatum:
And it's merged, thanks for your contribution!

PR #290, opened 2022-11-15
Test PR after master to main rename

gregtatum:


gregtatum:
CI is not running, but I'll wait a bit to see if maybe there is just a lag in the triggers.

gregtatum:
Saweet. It works meow 😸

PR #293, opened 2022-11-17
Test the intl-memoizer concurrent mode

gregtatum:
This is part of my effort to add to the test coverage for the project.

### DIFF
--- a/intl-memoizer/src/lib.rs
+++ b/intl-memoizer/src/lib.rs
@@ -432,22 +432,4 @@ mod tests {
             assert_eq!(result, Ok(PluralCategory::OTHER));
         }
     }
-
-    #[test]
-    fn test_debug_concurrent() {
-        let memoizer = concurrent::IntlLangMemoizer::new("en".parse().unwrap());
-        assert!(
-            format!("{:?}", memoizer).starts_with("IntlLangMemoizer"),
-            "The memoizer debug prints."
-        );
-    }
-
-    #[test]
-    fn test_debug_single_thread() {
-        let memoizer = IntlLangMemoizer::new("en".parse().unwrap());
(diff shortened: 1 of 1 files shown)


intl-memoizer/src/lib.rs
```diff
@@ -407,4 +408,46 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_debug_single_thread() {
+        let memoizer = IntlLangMemoizer::new("en".parse().unwrap());
+        assert!(
+            format!("{:?}", memoizer).starts_with("IntlLangMemoizer"),
+            "The memoizer debug prints."
+        );
+    }
```
nordzilla, line 452: > thought

I'm a bit unclear on what this is testing that is of use. Any object that derives `Debug` will print _something_ that contains the object's type name.

nordzilla (approved):
This looks good to me, though I don't understand the purpose of the debug tests.

intl-memoizer/src/lib.rs
gregtatum: I was dogmatically testing anything that was uncovered, which the Debug trait was being listed as uncovered. Maybe it's not the best use of a test.

nordzilla: Hmm. My preference would be if there was a way to ignore `Debug` trait from the coverage report. I don't know if we have that ability, but I would like that more than having tests that don't really test anything substantial just for the sake of having coverage here. It's not the end of the world to keep them, but I don't think we gain anything from this other than a higher number on the report.

gregtatum: Weirdly enough, the coverage isn't even showing up as missing there now.

PR #294, opened 2022-11-17
Add test coverage for fluent-resmgr

gregtatum:
This component doesn't really seem production ready yet, but this adds some tests. You will notice I reference some bugs here that I filed after adding some tests.

### DIFF
--- a/fluent-resmgr/src/resource_manager.rs
+++ b/fluent-resmgr/src/resource_manager.rs
@@ -15,12 +15,30 @@ fn read_file(path: &str) -> Result<String, io::Error> {
     fs::read_to_string(path)
 }
 
+/// [ResourceManager] provides a standalone solution for managing localization resources which
+/// can be used by `fluent-fallback` or other higher level bindings.
 pub struct ResourceManager {
     resources: FrozenMap<String, Box<FluentResource>>,
     path_scheme: String,
 }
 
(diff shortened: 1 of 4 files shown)


fluent-resmgr/src/resource_manager.rs
```diff
@@ -138,3 +172,72 @@ impl BundleGenerator for ResourceManager {
...
+    fn get_bundle_error() {
+        let res_mgr = ResourceManager::new("./tests/resources/{locale}/{res_id}".into());
+        let _bundle = res_mgr.get_bundle(vec![langid!("en-US")], vec!["nonexistent.ftl".into()]);
+    }
+
+    // TODO - Syntax errors should be surfaced.
+    // https://github.com/projectfluent/fluent-rs/issues/280
+    #[test]
+    fn get_bundle_ignores_errors() {
+        // Note that this is the behavior at the time of the writing of this test.
```
nordzilla, line 222: > thought

As a reader I'm not clear on what this comment is trying to express. 

"this is the behavior" doesn't mean very much to me. I wouldn't necessarily know if the behavior was updated and this comment was neglected to be removed.

nordzilla (approved):
These tests look good to me, along with the renames. 

I just had one comment about a comment that I didn't understand very well.

Issue #287, opened 2022-11-10
Clippy should be run in CI


gregtatum:
We should add it to a workflow in: `.github/workflows`

alerque:
This was implemented in #342. I might submit some tweaks to CI usage across the board but this has been implemented.

PR #295, opened 2022-11-18
Clippy fixes

gybrish:
Fixing up clippy warnings/errors before we look to address #287

### DIFF
--- a/fluent-bundle/tests/helpers/mod.rs
+++ b/fluent-bundle/tests/helpers/mod.rs
@@ -88,7 +88,7 @@ pub fn get_defaults(path: &str) -> Result<TestDefaults, io::Error> {
     Ok(serde_yaml::from_str(&s).expect("Parsing YAML failed."))
 }
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct TestBundle {
     pub name: Option<String>,
@@ -102,7 +102,7 @@ pub struct TestBundle {
     pub errors: Vec<TestError>,
 }
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct TestResource {
     pub name: Option<String>,
@@ -120,7 +120,7 @@ pub struct TestSetup {
     pub resources: Vec<TestResource>,
 }
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct TestError {
     #[serde(rename = "type")]
@@ -186,7 +186,7 @@ pub struct TestFixture {
     pub suites: Vec<TestSuite>,
 }
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct BundleDefaults {
     #[serde(rename = "useIsolating")]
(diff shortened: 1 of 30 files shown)


gybrish:
Not sure why the commits for the rust fmt workflow are included in this, let me know if I can fix it up somehow.

gregtatum:
It looks like you merged rather than fast forwarded: https://code-maven.com/enforce-fast-forward-as-merge-strategy

I suspect you ran into issues because I recently changed the `master` branch to `main`, so you will probably need to change it locally a well. You can then pull the latest changes and rebase your work.

You could also fetch the latest changes, and then reset --hard to `main` when you are in your `clippy_fixes` branch and then `git cherry-pick cf5aa9e` since your work is a single commit.

fluent-bundle/tests/helpers/mod.rs
```diff
@@ -88,7 +88,7 @@ pub fn get_defaults(path: &str) -> Result<TestDefaults, io::Error> {
     Ok(serde_yaml::from_str(&s).expect("Parsing YAML failed."))
 }
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
```
gregtatum, line 91: Question: Which clippy rule is this? It seems fine since these are in tests.

fluent-fallback/src/errors.rs
```diff
@@ -2,7 +2,7 @@ use fluent_bundle::FluentError;
 use std::error::Error;
 use unic_langid::LanguageIdentifier;
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
```
gregtatum, line 5: Question (blocking): Similar to above, I'd like to know what this clippy rule is before merging, since it's changing a public facing API.

fluent-fallback/src/types.rs
```diff
@@ -53,7 +53,7 @@ pub enum ResourceType {
 }
 
 /// A resource identifier for a localization resource.
-#[derive(Debug, Clone, Hash)]
```
gregtatum, line 56: Question (blocking): Which clippy rule is this? I don't understand this change (probably since I don't know the clippy rule)

fluent-syntax/src/ast/helper.rs
```diff
@@ -5,15 +5,15 @@ use super::Comment;
 // This is a helper struct used to properly deserialize referential
 // JSON comments which are single continous String, into a vec of
 // content slices.
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone)]
```
gregtatum, line 8: TODO (me): Review this and any other `Eq` traits after the above questions are answered.

gregtatum (changes requested):
Thanks for the PR! Overall this looks good. I'm marking as request changes for:

* Clean commit history without extraneous commits.
* I have a couple of blocking questions I'd like answered before I re-review.

After addressing those this looks mergeable to me!

fluent-bundle/tests/helpers/mod.rs
gybrish: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

For tests yeah, I'd say this rule is not applicable, we can suppress the warning

fluent-fallback/src/errors.rs
gybrish: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

This is more of a judgement call tbh, is there a reason to limit it to `PartialEq` and not `Eq`?

fluent-fallback/src/types.rs
gybrish: https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq

There's an `impl PartialEq` which only uses `self.value` for equality testing, while the derived `Hash` would use both `value` & `resource_type`, which breaks the invariant `k1 == k2 => hash(k1) == hash(k2)` (from the clippy rule)

fluent-syntax/src/ast/helper.rs
gybrish: Just for reference: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

gybrish:
Looks like it's difficult to sync the fork after the branch name change - will resubmit a new PR.

PR #296, opened 2022-11-19
Fixed up clippy warnrings/errors

gybrish:
Fixing up clippy warnings/errors before working on #287 

For reference:
`PartialEq, Eq`: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq
Derived `Hash` with `impl PartialEq`: https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq

### DIFF
--- a/fluent-fallback/src/types.rs
+++ b/fluent-fallback/src/types.rs
@@ -53,7 +53,7 @@ pub enum ResourceType {
 }
 
 /// A resource identifier for a localization resource.
-#[derive(Debug, Clone, Hash)]
+#[derive(Debug, Clone)]
 pub struct ResourceId {
     /// The resource identifier.
     pub value: String,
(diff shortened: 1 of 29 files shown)


fluent-fallback/src/types.rs
```diff
@@ -120,6 +120,12 @@ impl PartialEq for ResourceId {
     }
 }
 
+impl std::hash::Hash for ResourceId {
```
gregtatum, line 123: Thought (no action needed): Ah, I'm trying to reason through this, but it seems like we may have some correctness issues around handling multiple resources that have different optional values. Previously the behavior looks like you could add multiple `ResourceId` values with differing `ResourceType`s, but now it will only have one. I'm not sure what the correct resolution is here, or if anything needs to happen. Maybe I should add some tests around this area.

This seems like a low-probability edge case, but should probably be reasoned through.

gregtatum (approved):
This all looks good to me. Thanks for the contribution!

PR #297, opened 2022-11-30
Start using workspace dependencies

adriantombu:
As discussed in https://github.com/projectfluent/fluent-rs/pull/282#discussion_r1035019918 here's a separate PR to test the ability to use workspace dependencies that were introduced in Rust 1.64.

From what I see in the [Firefox Rust update policy](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html) it should be ok as this is the Rust version used in the latest stable release of Firefox (lucky us haha).

Screenshot 2022-11-30 at 12 46 55

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,20 @@ members = [
 exclude = [
     "fluent-cli",
 ]
+
+[workspace.dependencies]
+criterion = "0.3"
+fluent-langneg = "0.13"
+futures = "0.3"
+iai = "0.1"
+intl_pluralrules = "7.0.1"
+rustc-hash = "1"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+thiserror = "1.0"
+tokio = "1.0"
+unic-langid = "0.9"
+
+fluent-bundle = { version = "0.15.2", path = "fluent-bundle" }
+fluent-fallback = { version = "0.7.0", path = "fluent-fallback" }
+fluent-syntax = { version = "0.11", path = "fluent-syntax" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ futures = "0.3"
 iai = "0.1"
 intl_pluralrules = "7.0.1"
 rustc-hash = "1"
-serde = { version = "1.0", features = ["derive"] }
+serde = "1.0"
 serde_json = "1.0"
 thiserror = "1.0"
 tokio = "1.0"
@@ -29,4 +29,5 @@ unic-langid = "0.9"
 
 fluent-bundle = { version = "0.15.2", path = "fluent-bundle" }
 fluent-fallback = { version = "0.7.0", path = "fluent-fallback" }
+fluent-pseudo = { version = "0.3", path = "fluent-pseudo" }
 fluent-syntax = { version = "0.11", path = "fluent-syntax" }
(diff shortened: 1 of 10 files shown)


gregtatum:
This also needs testing with `./mach vendor rust` on the Firefox codebase. I think you can run it locally by replacing the paths to a local repo, and then running it. Our vendor script uses a custom python vendoring system that may not yet know about what `cargo` and `rustc` can do yet.

https://firefox-source-docs.mozilla.org/build/buildsystem/rust.html

Would you be interested in taking a stab at this @adriantombu or would you prefer me or @nordzilla to try it. It may be a bit before we can get to it if it's us.

Cargo.toml
```diff
@@ -13,3 +13,19 @@ members = [
...
 ]
+
+[workspace.dependencies]
+criterion = "0.3"
+fluent-langneg = "0.13"
+futures = "0.3"
+iai = "0.1"
+intl_pluralrules = "7.0.1"
+rustc-hash = "1"
+serde = { version = "1.0", features = ["derive"] }
```
gregtatum, line 24: Suggestion: I don't think `features` should be enabled/disabled at the workspace level, they should be done at the crate level.

gregtatum (changes requested):
The code itself looks good to me. I'm marking this as "request changes" only so that the status is clear that we need to block on manually testing the vendor process in Firefox. I left details in comments on what this process should look like.

adriantombu:
> Would you be interested in taking a stab at this @adriantombu or would you prefer me or @nordzilla to try it. It may be a bit before we can get to it if it's us.

@gregtatum I can have a look at that, I have some free time currently 😁

Cargo.toml
```diff
@@ -13,3 +13,19 @@ members = [
 exclude = [
     "fluent-cli",
 ]
+
+[workspace.dependencies]
+criterion = "0.3"
+fluent-langneg = "0.13"
```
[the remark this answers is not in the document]
zbraniecki: `fluent-langneg` should migrate to icu4x. But I'm not confident that icu4x should be providing language negotiation.

In terms of moving that into this repo - sure!

adriantombu:
I tested against the Firefox codebase locally successfully and added a [documentation file](https://github.com/adriantombu/fluent-rs/blob/task/workspace-dependencies/docs/local-testing-firefox.md) to explain the different steps to achieve it.

Cargo.toml
```diff
@@ -13,3 +13,21 @@ members = [
...
+iai = "0.1"
+intl_pluralrules = "7.0.1"
+rustc-hash = "1"
+serde = "1.0"
+serde_json = "1.0"
+thiserror = "1.0"
+tokio = "1.0"
+unic-langid = "0.9"
+
+fluent-bundle = { version = "0.15.2", path = "fluent-bundle" }
```
gregtatum, line 30: Question: Do you know if this works without the version, or if `cargo-release` automatically updates this version number? It would be nice not to have to maintain multiple sets of version numbers, but I'm unsure how this works.

adriantombu: I just made a quick test with `cargo-release` and the main `Cargo.toml` is indeed updated. But it probably doesn't make much sense to specify the version there so I'll clean that ^^

nordzilla (approved):
LGTM, especially after the most recent commit regarding the version numbers. Thanks so much!

gregtatum:
@adriantombu Thanks once again for the contribution! 🎉

Issue #238, opened 2021-09-14
Concurrent FluentBundle not accessible


FKanzler:
Hello,

unfortunately I am not able to figure out how to access the concurrent specialization of FluentBundle.

In the documentation it says:
> ### Concurrency
>---
> As you may have noticed, fluent_bundle::FluentBundle is a specialization of fluent_bundle::bundle::FluentBundle which works with an IntlLangMemoizer over RefCell. In scenarios where the memoizer must work concurrently, there’s an implementation of IntlLangMemoizer that uses Mutex and there’s FluentBundle::new_concurrent which works with that.

Simplified code:
```rust
use fluent_bundle::{FluentResource, FluentBundle}; //version: 0.15.1

let mut bundle = FluentBundle::new_concurrent(vec![langid!("en-US")]);
```
 
 Compiler error:
```
error[E0599]: no function or associated item named `new_concurrent` found for struct `FluentBundle<_, intl_memoizer::IntlLangMemoizer>` in the current scope
   --> libs\translate\src\lib.rs:139:52
    |
139 |                     let mut bundle = FluentBundle::new_concurrent(vec![langid!("en-US")]);
    |                                                    ^^^^^^^^^^^^^^ function or associated item not found in `FluentBundle<_, intl_memoizer::IntlLangMemoizer>`

error: aborting due to previous error
```

Is there something I'm missing here? How could this be made to work?

gregtatum:
The work here is to re-export the `new_concurrent` as suggested above. The IntlLangMemoizer is covered by #256.

adriantombu:
@gregtatum From what I understand the problem comes from [here](https://github.com/projectfluent/fluent-rs/blob/main/fluent-bundle/src/lib.rs#L122) where we redefine `FluentBundle` as

```pub type FluentBundle<R> = bundle::FluentBundle<R, intl_memoizer::IntlLangMemoizer>```

thus prohibiting any use of the `new_concurrent` function with a `use fluent_bundle::{FluentBundle}` import.

---

A simple solution can be to create the following type

```pub type FluentBundleConcurrent<R> = concurrent::FluentBundle<R, concurrent::IntlLangMemoizer>;```

to allow `use fluent_bundle::{FluentBundleConcurrent}`. What are your thoughts on this?

alerque:
[...]

PR #300, opened 2022-12-13
Export a new type FluentBundleConcurrent

adriantombu:
Fixes https://github.com/projectfluent/fluent-rs/issues/238


zbraniecki:
I think I'd prefer `fluent_bundle::concurrent::FluentBundle` rather than `fluent_bundle::FluentBundleConcurrent`.

My main reason is that I doubt people will be using both, concurrent and non-concurrent, in the same module, so really all you need is to import the correct one.

That's what we do with `Memoizable` and `IntlMemoizer` and I'd prefer to continue that here.

adriantombu:
Makes sense, I'll just close this PR then ^^

timthelion:
That's how things were back in 0.14.4 https://docs.rs/fluent/0.14.4/fluent/concurrent/type.FluentBundle.html not sure how the regression happened.

PR #301, opened 2022-12-14
Add FluentValue.into_string to prevent String clone

JasperDeSutter:
This method works the same as FluentValue.as_string, but takes self by-value to prevent cloning Strings. All usages of as_string have been replaced by into_string (FluentValues are always dropped right after converting to string).

### DIFF
--- a/fluent-bundle/src/types/mod.rs
+++ b/fluent-bundle/src/types/mod.rs
@@ -235,6 +235,9 @@ impl<'source> FluentValue<'source> {
     }
 
     /// Converts the [`FluentValue`] to a string.
+    ///
+    /// Clones inner values when owned, borrowed data is not cloned.
+    /// Prefer using [`FluentValue::into_string()`] when possible.
     pub fn as_string<R: Borrow<FluentResource>, M>(&self, scope: &Scope<R, M>) -> Cow<'source, str>
     where
         M: MemoizerKind,
@@ -254,6 +257,9 @@ impl<'source> FluentValue<'source> {
     }
 
     /// Converts the [`FluentValue`] to a string.
+    ///
+    /// Takes self by-value to be able to skip expensive clones.
+    /// Prefer this method over [`FluentValue::as_string()`] when possible.
     pub fn into_string<R: Borrow<FluentResource>, M>(self, scope: &Scope<R, M>) -> Cow<'source, str>
     where
         M: MemoizerKind,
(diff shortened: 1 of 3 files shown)


zbraniecki:
Thank you! That's an exciting per optimization!

fluent-bundle/src/types/mod.rs
```diff
@@ -253,6 +253,25 @@ impl<'source> FluentValue<'source> {
         }
     }
 
+    /// Converts the [`FluentValue`] to a string.
```
gregtatum, line 256: Suggestion (docs): The docs here are identical to `as_string` and it took some detailed reading to understand the difference. Could you update the code comment here and above in `as_string` explaining the difference for the uses?

The key difference is that `into_string` takes ownership of `self`.

fluent-bundle/src/types/mod.rs
```diff
@@ -253,6 +253,25 @@ impl<'source> FluentValue<'source> {
...
+    where
+        M: MemoizerKind,
+    {
+        if let Some(formatter) = &scope.bundle.formatter {
+            if let Some(val) = formatter(&self, &scope.bundle.intls) {
+                return val.into();
+            }
+        }
+        match self {
+            FluentValue::String(s) => s,
```
gregtatum, line 267: Praise: That's a quite nice optimization.

gregtatum (changes requested):
Thanks for the PR! I have a small request for change, which is to clarify the docs a bit. I set the review as "request changes" so the status is clear. Please re-request review after you push new changes so that it will get back on my review queue.

I may be out a bit for the holidays depending on the timing, but I'll try and get to the response if I'm around.

fluent-bundle/src/types/mod.rs
JasperDeSutter: The docs should be a little better now. Thanks for the reviews btw!

fluent-bundle/src/types/mod.rs
```diff
@@ -254,6 +257,9 @@ impl<'source> FluentValue<'source> {
     }
 
     /// Converts the [`FluentValue`] to a string.
+    ///
+    /// Takes self by-value to be able to skip expensive clones.
+    /// Prefer this method over [`FluentValue::as_string()`] when possible.
```
gregtatum, line 262: Praise: Nice docs!

gregtatum (approved):
Thanks for the PR!

PR #302, opened 2022-12-14
Fix crash when parsing multiline CRLF comment

### LINKED ISSUE #220: Improper handling of Windows line feeds in FTL files
Hi, I'm working on porting Fluent.rs to .Net and I stumbled upon a possible error in the parser:

If you try to parse the following string `# msg\r\n# yes\r\n#\r\nfoo=Foo` you'll get errors. The same example on the [Fluent playground](https://www.projectfluent.org/play/), gets you:

```json
{
    "type": "Resource",
    "body": [
        {
            "type": "Message",
            "id": {
                "type": "Identifier",
                "name": "foo",
                "span": {
                    "type": "Span",
                    "start": 14,
                    "end": 17
                }
            },
            "value": {
                "type": "Pattern",
                "elements": [
                    {
                        "type": "TextElement",
                        "value": "Foo",
                        "span": {
                            "type": "Span",
                            "start": 20,
                            "end": 23
                        }
                    }
                ],
                "span": {
                    "type": "Span",
                    "start": 20,
                    "end": 23
                }
            },
            "attributes": [],
            "comment": {
                "type": "Comment",
                "content": "msg\nyes\n",
                "span": {
                    "type": "Span",
                    "start": 0,
                    "end": 13
                }
            },
            "span": {
                "type": "Span",
                "start": 0,
                "end": 23
            }
        }
    ],
    "span": {
        "type": "Span",
        "start": 0,
        "end": 24
    }
}
```

Which makes the most sense. I managed to track down the offending line to this branch:
https://github.com/projectfluent/fluent-rs/blob/4d13b24140409b377362442cf748a7f789c3f822/fluent-syntax/src/parser/comment.rs#L34

If the line was:
```rust
} else if self.is_current_byte(b'\n') 
  || (self.is_current_byte(b'\r') &&  self.is_byte_at(b'\n', self.ptr + 1) { 
```

I believe the error wouldn't happen, but I haven't tested it.

JasperDeSutter:
fixes #220

### DIFF
--- a/fluent-syntax/src/parser/comment.rs
+++ b/fluent-syntax/src/parser/comment.rs
@@ -31,7 +31,7 @@ where
 
             if self.ptr == self.length {
                 break;
-            } else if self.is_current_byte(b'\n') {
+            } else if self.is_eol() {
                 content.push(self.get_comment_line());
             } else {
                 if let Err(e) = self.expect_byte(b' ') {
@@ -52,7 +52,7 @@ where
 
     pub(super) fn skip_comment(&mut self) {
         loop {
-            while self.ptr < self.length && !self.is_current_byte(b'\n') {
+            while self.ptr < self.length && !self.is_eol() {
                 self.ptr += 1;
             }
             self.ptr += 1;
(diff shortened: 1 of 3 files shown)


gregtatum (approved):
Thanks for the fix! I confirmed the test crashes without your fix.

Issue #269, opened 2022-08-09
Number formatting isn't properly implemented


juliancoffee:
It seems that fluent-rs can parse options for numbers, but not use them.
Looking at FluentNumber::as_string(), the only option that gets used is minimum fraction digits.
```rust
    pub fn as_string(&self) -> Cow<'static, str> {
        let mut val = self.value.to_string();
        if let Some(minfd) = self.options.minimum_fraction_digits {
            if let Some(pos) = val.find('.') {
                let frac_num = val.len() - pos - 1;
                let missing = if frac_num > minfd {
                    0
                } else {
                    minfd - frac_num
                };
                val = format!("{}{}", val, "0".repeat(missing));
            } else {
                val = format!("{}.{}", val, "0".repeat(minfd));
            }
        }
        val.into()
    }
 ```
So, things to implement:
- [ ] Percent formatting
- [ ] Currency formatting
- [x] Decimal formatting (implicitly assumed, I guess)
- [ ] use_grouping
- [ ] minimum_integer_digits
- [x] minimum_fraction_digits
- [ ] maximum_fraction_digits
- [ ] minimum_significant_digits
- [ ] maximum_significant_digits

juliancoffee:
@zbraniecki 
I thought about implementing percentage formatting, but I'm a bit worried about language-dependent formatting, do you have any advice on how to do it?

zbraniecki:
I'd wait for ICU4X NumberFormatter, we have scaffolding for it in https://unicode-org.github.io/icu4x-docs/doc/fixed_decimal/index.html for some aspects, and https://github.com/unicode-org/icu4x/issues/1441 for others.

Maybe you'd like to help with that? :)

juliancoffee:
> I'd wait for ICU4X NumberFormatter, we have scaffolding for it in https://unicode-org.github.io/icu4x-docs/doc/fixed_decimal/index.html for some aspects, and [unicode-org/icu4x#1441](https://github.com/unicode-org/icu4x/issues/1441) for others.
> 
> Maybe you'd like to help with that? :)

Yeah, I was looking at that a bit. What are the approximate steps for working on that?

zbraniecki:
> What are the approximate steps for working on that?

You can go to issue #1441 and say you'd like to take it and ask for mentorship. Someone from our team will mentor you through that.

[...]

PR #303, opened 2022-12-15
ICU4x number formatting

JasperDeSutter:
Adds basic decimal formatting for FluentNumber.as_string(). Many features are still missing like currency symbols and percentage formatting. #269

Since data providers need to be managed explicitly, it needs to be handed to the memoizer to be able to construct a formatter. This required a new argument + associated type on Memoizable.

Something to figure out is where the data comes from. I've went with a simple FluentBundle.set_number_format_provider, which seems most flexible to me. I've added a test that uses icu_testdata for this, but other providers should work as well.
https://icu4x.unicode.org/doc/icu/#data-management

### DIFF
--- a/fluent-bundle/tests/types_test.rs
+++ b/fluent-bundle/tests/types_test.rs
@@ -126,7 +126,7 @@ fn fluent_number_style() {
     let num = FluentNumber::new(0.2, FluentNumberOptions::default());
     assert_eq!(num.as_string_basic(), "0.2");
 
-    let opts = FluentNumberOptions {
+    let mut opts = FluentNumberOptions {
         minimum_fraction_digits: Some(3),
         ..Default::default()
     };
@@ -134,8 +134,26 @@ fn fluent_number_style() {
     let num = FluentNumber::new(0.2, opts.clone());
     assert_eq!(num.as_string_basic(), "0.200");
 
-    let num = FluentNumber::new(2.0, opts);
+    let num = FluentNumber::new(2.0, opts.clone());
     assert_eq!(num.as_string_basic(), "2.000");
+
+    opts.minimum_integer_digits = Some(3);
+    opts.minimum_fraction_digits = None;
+
+    let num = FluentNumber::new(2.0, opts.clone());
+    assert_eq!(num.as_string_basic(), "002");
+
+    let num = FluentNumber::new(0.2, opts.clone());
+    assert_eq!(num.as_string_basic(), "000.2");
+
+    opts.minimum_integer_digits = None;
+    opts.maximum_significant_digits = Some(4);
+
+    let num = FluentNumber::new(12345.0, opts.clone());
+    assert_eq!(num.as_string_basic(), "12340");
+
+    let num = FluentNumber::new(1.2345, opts);
+    assert_eq!(num.as_string_basic(), "1.234");
 }
 
 #[test]
--- a/fluent-bundle/tests/types_test.rs
+++ b/fluent-bundle/tests/types_test.rs
@@ -1,6 +1,7 @@
 use fluent_bundle::resolver::Scope;
 use fluent_bundle::types::{
-    FluentNumber, FluentNumberCurrencyDisplayStyle, FluentNumberOptions, FluentNumberStyle, FluentNumberUseGrouping,
+    FluentNumber, FluentNumberCurrencyDisplayStyle, FluentNumberOptions, FluentNumberStyle,
+    FluentNumberUseGrouping,
 };
 use fluent_bundle::FluentArgs;
 use fluent_bundle::FluentBundle;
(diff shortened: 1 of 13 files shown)


gregtatum (commented):
Thought: I'm not done reviewing this, but figured I would leave some early feedback. The implementation is looking pretty good in my initial review. I still need to complete a more thorough one.

This is going to take some careful consideration to merge since it can impact binary sizes. In addition this requires the users to have a data provider, which could be pretty significant. I'm wondering if this feature might need to start out as a trait.

Firefox is really sensitive to binary size increases, so I wouldn't feel comfortable merging until I knew we could manage the dependency changes and the additional data requirements. If Firefox has these issues, then it's possible other implementations will as well. 

There are essentially 2 places new binary size can increase here, the dependency executable code, and in the data in the data provider. It would be good to measure both of these. I wonder if you just shipped with an `en-001` locale (or equivalent) as a single locale, if it would be the equivalent behavior as what's going on now in Firefox and the fluent-rs project.

(I'm not the original author of this code, so I'd have to read through the code to fully know the answer here.)

Do you have any thoughts in this space @JasperDeSutter?

JasperDeSutter:
I compared the average binary size of all examples in fluent-bundle: (stripped release build)

605KB - without decimal formatting (main branch): 100%
643KB - with number formatting code, no data: +6%
780KB - with decimal and fallback data for all locales: +29%
716KB - with icu_testdata (icu_decimal feature, only test locales): +18%

To me it looks like a good idea to not include data, but I'm not sure whether a feature flag is warranted for the formatting code.

gregtatum (changes requested):
> To me it looks like a good idea to not include data, but I'm not sure whether a feature flag is warranted for the formatting code.

I agree. 38kb is not too bad in terms of binary size increase. Thanks for quantifying it! That's really helpful. Firefox plans on eventually integrating ICU4X, but there's not a concrete written plan yet. The binary increase seems acceptable, and eventually Firefox will have its own data provider, and will control that data cost.

I'd like to get some feedback with other Mozillians in this space, and we're hitting the holidays where people are out.

I think the next steps here would be to add the trait in just for the data, and make sure the ICU4X bits are well documented with suggestions on the proper way to integrate the data. I might run this by for feedback from the ICU4X team after soliciting feedback from other Mozillians.

I'm marking request changes since there's actionable feedback already, feel free to re-request review if you would like a more thorough code review at this point before adding the trait in.

gregtatum:
> I'd like to get some feedback with other Mozillians in this space, and we're hitting the holidays where people are out.

Also to be clear here, I'd like to wait until after the holidays at least before merging since people are probably out since this affects the architecture of this project and Firefox. I think any code updates and reviews can happen before then, but I just wanted to set expectations.

gregtatum:
I'm closing as stale since there were requested changes. Feel free re-open.

fee1-dead:
any chance we could get a review on this?

alerque:
@fee1-dead Sure, go ahead and review. Test results welcome. Only half jesting, if you want to help move things along the best way to do that is contribute by reviewing and testing yourself and reporting results.

Also be aware that because this introduces a bunch of new dependencies we'll probably have to hold off on merging until we can base it on or at least compare it with what is happening in #335.

fluent-bundle/tests/types_test.rs
```diff
@@ -154,3 +173,25 @@ fn fluent_number_to_operands() {
         }
     );
 }
+
+#[test]
+fn fluent_number_grouping() {
+    let langid_ars = langid!("ccp");
+    let mut bundle: FluentBundle<FluentResource> = FluentBundle::new(vec![langid_ars]);
+    #[allow(deprecated)]
+    bundle.set_icu_data_provider(Box::new(icu_testdata::any_no_fallback()));
```
fee1-dead, line 182: This is deprecated, and it tells us to use `compiled_data` instead.

From a first look, you probably need a new type that implements [`DynamicDataProvider`](https://docs.rs/icu_provider/1.4.0/icu_provider/trait.DynamicDataProvider.html) using [`icu_decimal::provider::Baked`](https://docs.rs/icu_decimal/latest/icu_decimal/provider/struct.Baked.html) by asserting the data key [passed in](https://docs.rs/icu_provider/1.4.0/icu_provider/trait.DynamicDataProvider.html) is equal to [`DecimalSymbolsV1Marker`](https://docs.rs/icu_decimal/latest/icu_decimal/provider/struct.DecimalSymbolsV1Marker.html#impl-KeyedDataMarker-for-DecimalSymbolsV1Marker).

Then you can use [`as_any_provider`](https://docs.rs/icu_provider/latest/icu_provider/any/trait.AsDynamicDataProviderAnyMarkerWrap.html#tymethod.as_any_provider) to convert that to the provider this needs.

JasperDeSutter: Thanks, I've added a baked data provider that can be used both from tests/examples and by users of the library.

fluent-bundle/src/icu_data_provider.rs
```diff
@@ -0,0 +1,26 @@
...
+/// Example:
+/// ```
+/// use fluent_bundle::{FluentBundle, FluentIcuDataProvider, FluentResource};
+/// use unic_langid::langid;
+///
+/// let langid = langid!("en-US");
+/// let mut bundle: FluentBundle<FluentResource> = FluentBundle::new(vec![langid]);
+/// bundle.set_icu_data_provider(Box::new(FluentIcuDataProvider));
+/// ```
+pub use BakedDataProvider as FluentIcuDataProvider;
```
fee1-dead, line 26: It's possible that this file can increase compile times by a bit. Probably best to include a feature flag that can turn the built in provider off?

JasperDeSutter: That makes sense to me, though I'd like to descope that as this PR is already hard to get merged as-is and there's still more experimentation needed to figure out the whole ICU data story.

fee1-dead (approved):
LGTM.

alerque:
@fee1-dead Did you review this in context with other ICU4x related work such as in #335 and the discussions in a number of related issues. We could use help reviewing and sorting out all the moving pieces, but just saying this one PR "lgtm" in isolation doesn't move the ball forward much.

fee1-dead:
if I actually do get time I can review the other PR, but at a first glance this PR is isolated enough (only uses icu4x for number formatting) I don't think it would be very probable for a conflict with that effort to use icu4x in more places.

PR #304, opened 2022-12-16
Allow customizing argument resolver

JasperDeSutter:
The main value add for a custom argument resolver is not needing to put arguments into a FluentArgs, which has no type safety. See the typesafe_messages example that demonstrates this use case.
The first commit optimizes FluentArg by not storing keys as Cows, but as regular &str instead. This reduces a small overhead in data size and arg lookup performance.


fluent-bundle/src/resolver/scope.rs
```diff
@@ -138,3 +145,14 @@ impl<'bundle, 'ast, 'args, 'errors, R, M> Scope<'bundle, 'ast, 'args, 'errors, R
         }
     }
 }
+
+pub trait ArgumentResolver<'a>: Copy {
```
gregtatum, line 149: Suggestion: Please add docs on what this is and how to use it.

fluent-bundle/examples/typesafe_messages.rs
```diff
@@ -0,0 +1,129 @@
+use std::borrow::Cow;
```
gregtatum, line 1: Suggestion: Please add a description here at the top of what this is demonstrating, so that users can quickly gain context.

fluent-bundle/src/resolver/scope.rs
gregtatum: Also mentioning and linking to the `typesafe_messages` example would be useful.

gregtatum (approved):
This all seems reasonable to me, and allows for more programmatic integration with a Rust codebase. Since this is a design change I would like to get another reviewer on this to double check.

@nordzilla Would you mind looking at this briefly? The last commit includes an example that shows how to use it which I would appreciate your opinion on. You can also see the implementation of `ArgumentResolver` in `fluent-bundle/src/resolver/scope.rs`. You can track down the calls of it by searching for `.resolve`. I don't think you need to look at the first commit, since that's more of an implementation detail.

gregtatum:
Oh and also thanks for the PR :)

alerque (approved):
I rebased this just to fix the commit messages and move the doc fixup commit into the one it was fixing. I did not resolve the merge conflict that has come up since this was contributed. @JasperDeSutter is there any chance you are still around and could have a look at this? This project is in a little better position to move on with contributions now (see #347 for background).

alerque (changes requested):
I rebased this just to fix the commit messages and move the doc fixup commit into the one it was fixing. I did not resolve the merge conflict that has come up since this was contributed. @JasperDeSutter is there any chance you are still around and could have a look at this? This project is in a little better position to move on with contributions now (see https://github.com/projectfluent/fluent-rs/issues/347 for background).

JasperDeSutter:
Great to see this project more alive again! I'll pick up my PRs that are still open, though it has been a while so it might take some time to get them all updated.

PR #305, opened 2022-12-21
Resolve more pattern types into &str references

JasperDeSutter:
The most important change is in `fluent-bundle/src/resolver/pattern.rs`, the rest is adding missing codepaths that were previously only done through `WriteValue` and are now also needed in `ResolveValue`.

There are more cases where we can resolve to a borrowed str than just the TextElement case. For example there might be a select that resolves into a single TextElement, or a TermReference which is a simple string.
With these changes, `ResolveValue` methods will be called until a pattern with more than 1 element is found, which needs string concatenation and is handled by the `WriteValue` methods as before.

When resolving into a `FluentValue::Error`, the `WriteValue` is called to get "{error}" formatting instead of an empty string.


JasperDeSutter:
Before these changes, Cow::Borrowed was returned 14 out of 174 times in fluent-bundle test cases. After the changes, I'm seeing 67 out of 174.

gregtatum:
@JasperDeSutter I didn't get a chance to finish reviewing all of your PRs before the holidays. I'll be back in the new year.

gregtatum:
I'm back from the holidays, but may still need a few days to get caught up for reviews.

fluent-bundle/src/resolver/inline_expression.rs
```diff
@@ -183,14 +183,72 @@ impl<'bundle> ResolveValue<'bundle> for ast::InlineExpression<&'bundle str> {
...
+                let result = scope
+                    .bundle
+                    .get_entry_term(id.name)
+                    .and_then(|term| {
+                        if let Some(attr) = attribute {
+                            term.attributes.iter().find_map(|a| {
+                                if a.id.name == attr.name {
+                                    Some(scope.track_resolve(&a.value))
+                                } else {
+                                    None
```
gregtatum, line 237: Suggestion (optional): This `None` case isn't covered by a test. It would be nice to have a test fixture somewhere with multiple terms in a single message that get resolved.

image

fluent-bundle/src/resolver/pattern.rs
```diff
@@ -91,13 +91,33 @@ impl<'bundle> ResolveValue<'bundle> for ast::Pattern<&'bundle str> {
     {
         let len = self.elements.len();
 
+        // more than 1 element means concatenation, which is more efficient to write to a String
+        // 1 element often means just a message reference that can be passed back as a Cow::Borrowed
         if len == 1 {
```
gregtatum, line 96: Suggestion: I would invert this statement to be about the case below it. Something like:

> If there is only 1 element, than it is more efficient to attempt to resolve as a message reference that can be passed back as a Cow::Borrowed rather than writing a new String.

As it is, it's a little confusing to read, since it is more about the case _below_ the if statement.

fluent-bundle/src/resolver/scope.rs
```diff
@@ -102,6 +102,49 @@ impl<'bundle, 'ast, 'args, 'errors, R, M> Scope<'bundle, 'ast, 'args, 'errors, R
...
+        }
+        let res = exp.resolve(self);
+        if self.dirty {
+            FluentValue::Error
+        } else {
+            res
+        }
+    }
+
+    /// Similar to [`Scope::track`], but resolves to a value
```
gregtatum, line 127: Suggestion (optional): This is referencing `Scope::track`, but `Scope::track` doesn't have any documentation. Would you mind writing a short comment above `Scope::track`, otherwise the comment makes it a little hard to see what's going on.

gregtatum (changes requested):
Thanks for the PR! I am only marking as "Request Changes" so that I can see your response to my feedback. The PR looks good, I had a few minor comments and suggestions. Some of them are optional, which if you don't have the time, I can merge without them.

Please re-flag me for review, and we can get this merged!

fluent-bundle/src/resolver/expression.rs
```diff
@@ -64,3 +64,45 @@ impl<'bundle> WriteValue<'bundle> for ast::Expression<&'bundle str> {
...
+                    }
+                    _ => {}
+                }
+
+                for variant in variants {
+                    if variant.default {
+                        return variant.value.resolve(scope);
+                    }
+                }
+                scope.add_error(ResolverError::MissingDefault);
```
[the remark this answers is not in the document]
JasperDeSutter: This is unreachable code since a default variant is required, which is checked by the parser already. I think this doesn't warrant an explicit resolver error since we know it wont ever arise. A panic would be better suited IMO.

alerque:
Rebased to update commit messages, get current CI jobs to run, and run `cargo clippy` on each commit. Working on rebase to *main* now.

alerque:
...and also fixed the `rustfmt` issue.

alerque (changes requested):
Thanks for the contribution(s).

It looks like this could use a few new unit tests.

fluent-bundle/src/resolver/inline_expression.rs
JasperDeSutter: Added a test fixture to cover this

fluent-bundle/src/resolver/scope.rs
JasperDeSutter: The Scope::track function now has a doc comment as well.

PR #306, opened 2022-12-21
Fix FluentValue::try_number accepting numbers

JasperDeSutter:
This seems to be a mistake since the documentation only describes its use for parsing strings.
By only accepting &str, there's no need for the to_string() call, making it more efficient. (AsRef<str> is also an option)
Note that this would be a breaking change since it's a public API.

### DIFF
--- a/fluent-bundle/src/args.rs
+++ b/fluent-bundle/src/args.rs
@@ -146,7 +146,7 @@ mod tests {
             args.get("name"),
             Some(&FluentValue::String(Cow::Borrowed("John")))
         );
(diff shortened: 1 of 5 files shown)


gregtatum (approved):
Your reasoning all seems good to me!

> This seems to be a mistake since the documentation only describes its use for parsing strings.

I'm the `git blame` for that documentation so I wrote it based on just reading the code and not talking to the original author 😝

It seems that if you wanted an API that took a number type, it would be infallible anyway. The fallible `try_` version would only be for using strings, so this API change seems reasonable to me, and a nice little optimization.

PR #309, opened 2022-12-22
Migrate to pin_cell crate

zbraniecki:
Now that https://github.com/withoutboats/pin-cell/issues/6 is resolved, we can migrate to `pin_cell`.


fluent-fallback/src/pin_cell/mod.rs
```diff
@@ -1,97 +0,0 @@
-#![deny(missing_docs, missing_debug_implementations)]
```
gregtatum, line 1: Thought: Looks like this will be another vendored crate for Firefox, but it should be direct enough to certify on the next vendoring.

gregtatum (approved):

gregtatum:
Uh oh, this has conflicts that need resolving.

fluent-fallback/src/pin_cell/mod.rs
zbraniecki: My logic is that we're not experts to maintain it in-house, so better to vendor no-dep utility

zbraniecki:
rebased

PR #310, opened 2022-12-22
Merge ResolveValue and WriteValue traits

JasperDeSutter:
This is a refactor with the goal to have only 1 implementation of Read/Resolve for ast expressions.
In #305 I had to duplicate some Write implementations to support Resolve as well. That's not necessary anymore after this PR.

The last commit reduces the number of lifetimes on the Scope struct so that the resolve code becomes a little nicer to work with. They didn't need to be separate as it doesn't affect public API.

I'm not sure what to name this, so suggestions to improve "WriteOrResolve" are welcome.

### DIFF
--- a/fluent-bundle/src/bundle.rs
+++ b/fluent-bundle/src/bundle.rs
@@ -21,7 +21,8 @@ use crate::entry::GetEntry;
 use crate::errors::{EntryKind, FluentError};
 use crate::memoizer::MemoizerKind;
 use crate::message::FluentMessage;
-use crate::resolver::{ResolveValue, Scope, WriteValue};
+use crate::resolver::pattern::{resolve_pattern, write_pattern};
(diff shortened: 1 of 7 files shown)


alerque:
Rebased on *main* to update commit messages and so we can see CI results. I haven't yet given this a serious review to consider what the implementations would be of the breaking changes. I also don't think we have a clear roadmap for upcoming releases but so far in the rush of stuff after the safe-harbor release we haven't implemented anything breaking yet, so we should consider whether we want to do a release series with only non-breaking changes before we move forward on any breaking ones.

alerque:
Rebased again this time using `cargo clippy --fix` on each commit, just because that's low hanging fruit to get our CI tests showing useful results. That actually made the last commit in the original series a `noop` because clippy fixed the same stuff in the second commit.

Issue #299, opened 2022-12-06
error[E0277]: `(dyn Any + 'static)` cannot be sent between threads safely


Sedorikku1949:
Hi, I know this error is an issue that has been closed, but I don't understand HOW to access the "new_concurrent" method.

When I use this code:
```rs
  {
    let ftl_string = include_str!("../assets/languages/fr.ftl");
    let res = FluentResource::try_new(ftl_string.to_string()).expect("Failed to parse an FTL string.");
    
    let lang = langid!("fr-FR");
    let mut bundle = FluentBundle::new(vec![lang.clone()]);

    bundle.add_resource(res).expect("Failed to add FTL resources to the bundle.");

    langs.insert(lang, bundle);
  }
```

It doesn't give any error (in that part, I use threads so this part broke my whole project).
But, when I try to use the `FluentBundle::new_concurrent` method instead of `FluentBundle::new`, it says that this function doesn't exist, and I can't find anything about it in the documentation.

How can we access to this method ?

zbraniecki:
Ah, good call. It's not very ergonomic now, you need to do:

```rust
let bundle = bundle::FluentBundle<_, intl_memoizer::concurrent::IntlMemoizer>;
````

**Comment by Sedorikku1949** [comment]
Thank you, it works perfectly now !

**Comment by Sedorikku1949** [comment]
For those who were also wondering how to do it:
```rs
// an example of a basic fluent implementation with a HashMap to store languages (works with threads!)
use std::collections::HashMap;
use fluent::{ FluentResource, bundle::FluentBundle };
use unic_langid::{self, LanguageIdentifier, langid};
use intl_memoizer;

// Very important !!!!
type TranslationType = FluentBundle<FluentResource, intl_memoizer::concurrent::IntlLangMemoizer>;

fn load() -> HashMap<LanguageIdentifier, TranslationType> {
  let mut langs: HashMap<LanguageIdentifier, TranslationType> = HashMap::new();
  {
    let ftl_string = include_str!("../assets/languages/fr.ftl"); // you can use a &str or read a file instead of include_str !
    let res = FluentResource::try_new(ftl_string.to_string()).expect("Failed to parse an FTL string.");
    
    let lang = langid!("fr-FR");
    let mut bundle = FluentBundle::new_concurrent(vec![lang.clone()]);

    bundle.add_resource(res).expect("Failed to add FTL resources to the bundle.");

    langs.insert(lang, bundle);
  }
  langs
}
```

[...]

PR #311, opened 2023-01-01
Add type alias for concurrent `FluentBundle`

squ1dd13:
As per the comment in #299.

I've created this alias in the `concurrent` module. That seemed like the best place to put it imo, but in order to do that I had to make the module public. I hope that's not a problem.

Cheers.


gregtatum:
Thanks for the PR. I added myself as a reviewer, but I'm just getting back from the holidays so it may take me a few days to get to reviews.

gregtatum (approved):
Thanks for the PR. Looks good to me!

PR #312, opened 2023-02-08
Treat tab as text, not whitespace

eemeli:
See https://github.com/projectfluent/python-fluent/issues/173

Tabs were defined to be text, not whitespace, in https://github.com/projectfluent/fluent/pull/167, but this change was not reflected here.

### DIFF
--- a/fluent-syntax/src/parser/slice.rs
+++ b/fluent-syntax/src/parser/slice.rs
@@ -10,7 +10,9 @@ impl<'s> Slice<'s> for String {
     }
 
     fn trim(&mut self) {
-        *self = self.trim_end().to_string();
+        *self = self
+            .trim_end_matches(|c| c == ' ' || c == '\r' || c == '\n')
+            .to_string();
     }
 }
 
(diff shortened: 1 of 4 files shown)


gregtatum (approved):
Looks good to me with Zibi's suggestion.

eemeli:
Extracted the closure to a function. Do I understand right that this effectively becomes a part of the public API as well? Or is there a way to make it available from `src/parser/slice.rs` to `src/serializer.rs` without doing so?

zbraniecki:
As you did it now, yes, because lib.rs has `pub mod parser` and parser has `pub use slice::matches_fluent_ws`.

But you can limit it by using `pub(crate) fn matches_fluent_ws` making this function visible only internally to the crate.
Whether you want to or not is not clear to me.

On one hand, if this is Fluent WS, then having access to a test of it from `parser` is nice, on the other it is increasing the stable API surface and if someone really needs it they can readd it at very low cost.
My weak lean is to keep it private until someone asks for it.

zbraniecki (approved):

PR #316, opened 2023-04-09
fluent-bundle: Bump serde_yaml to 0.9 from 0.8

waywardmonkeys:
0.9 was released about 8 months ago. There were breaking changes listed at the time:

https://github.com/dtolnay/serde-yaml/releases/tag/0.9.0

It does build and pass tests though ...

### DIFF
--- a/fluent-bundle/CHANGELOG.md
+++ b/fluent-bundle/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-  - …
+  - Bump `serde_yaml` to 0.9.
 
 ## fluent-bundle 0.15.2 (October 25, 2021)
   - Bump `self_cell` to 0.10.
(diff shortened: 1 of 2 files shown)


alerque:
The one concern that comes to mind off the top of my head is this bump includes a dependency bump for serde itself that goes past the release where the build process added a binary download (https://github.com/serde-rs/serde/issues/2538). I don't know what happened *after* that snafu so the version being used now might not have the same concern, but I would at least like to be sure we aren't introducing prebuilt binaries into the build process by making this bump.

alerque (approved):

alerque:
And ... since this was proposed `serde_yaml` has actually been archived by its original developer (with no notification and some drama) with no plans to turn it over to anybody to continue maintenance. Last I looked it wasn't clear what fork or alternative would be best to move to, but that was very early after the archive move. I'm sure eventually an alternative will be clear, and we should probably open an issue to track this.

In the mean time I still agree bumping to 0.9.x is the best immediate thing to do.

PR #320, opened 2023-07-23
ci: Stop using `actions-rs/toolchain`.

waywardmonkeys:
The `actions-rs` actions haven't been maintained in some time and result in a number of deprecation warnings from GitHub Actions.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -10,10 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Toolchain setup
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly
-          override: true
           components: rustfmt
 
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/format.yml
```diff
@@ -10,10 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Toolchain setup
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
```
flodolo, line 13: Should this be `@nightly` instead of `@master`?

flodolo (commented):
For what it's worth, plenty of projects seem to be switching to this GitHub action, unless they need support for `rust-toolchain.toml`
https://github.com/actions-rs/toolchain/issues/216

.github/workflows/format.yml
waywardmonkeys: Not needed since the tool chain is specified below. Did it that way since a component is specified.

flodolo (approved):

PR #325, opened 2023-08-31
Use "resolver = 2" in the virtual workspace.

waywardmonkeys:
This is the default for 2021 edition crates, but not for a virtual workspace, so it must be set explicitly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
     "fluent-syntax",
     "fluent-bundle",


xStrom (approved):

alerque (approved):

g2p (approved):

flodolo:
This seems safe, based on [this page](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html) and the setting in other Rust projects.

flodolo (approved):

PR #333, opened 2023-11-11
Add Capability to Localize Terms

HTGAzureX1212:
This PR adds a `get_term` method such that terms can also be localized via a similar API to that of messages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea
+
 /target/
 */target/
 **/*.rs.bk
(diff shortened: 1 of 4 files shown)


zbraniecki:
This is intentional. The hypothesis was that term is a private construct and should be changable without obsoleting.
If you want to use a term, construct a message that references it. 

@stasm to verify my words.

stasm:
That's correct. Quoting https://projectfluent.org/fluent/guide/terms.html:

> Terms are similar to regular messages but they can only be used as references in other messages. (...) The runtime cannot retrieve terms directly.

alerque:
As noted, terms are primarily useful specifically *because* they cannot be accessed by the runtime and thus are by definition up to the localizer. I appreciate this flexibility of having access to a way to compose messages out of smaller parts *if desired* without worrying about the code end of things and what usages I might be breaking. The list of messages is controlled by the code/consumer. Terms are controlled by the localizer and can/likely will be different between localizations. Adding this interface would compromise that and goes against why the spirit of why they are in the spec in the first place.

PR #342, opened 2024-02-14
Run clippy on CI

pitbuster:
First take on #287. I am not sure if running nightly `clippy` is a requirement for the project (in my experience it is quite unstable), but we can work on that if that's the case.

### DIFF
--- a/.github/workflows/clippy.yaml
+++ b/.github/workflows/clippy.yaml
@@ -13,10 +13,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Setup Rust toolchain and cache
-        uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
+      - name: Toolchain setup
+        uses: dtolnay/rust-toolchain@master
         with:
-          rustflags: ""
+          toolchain: stable
+          components: rustfmt
 
       - name: Clippy
         run: cargo clippy -- --all-features -Dwarnings
--- a/.github/workflows/clippy.yaml
+++ b/.github/workflows/clippy.yaml
@@ -1,0 +1,22 @@
+name: Clippy
+
+on:
+  pull_request:
+    branches: [ main ]
+  push:
+    branches: [ main ]
+
+
+jobs:
+  lints:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Rust toolchain and cache
+        uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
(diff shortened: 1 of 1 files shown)


.github/workflows/clippy.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+
+jobs:
+  lints:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Rust toolchain and cache
+        uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
```
waywardmonkeys, line 17: If this is added, it should use the same stuff for managing the toolchain as what's in the other CI jobs.

.github/workflows/clippy.yaml
```diff
@@ -0,0 +1,22 @@
...
+        with:
+          rustflags: ""
+
+      - name: Clippy
+        run: cargo clippy -- --all-features -Dwarnings
```
waywardmonkeys, line 22: This should probably be checking all targets in the workspace.

waywardmonkeys (changes requested):
I'm not sure if this should be a new job or just part of the existing test workflow.

.github/workflows/clippy.yaml
pitbuster: It should already be doing that.

alerque:
Personally (having setup and managed a lot of CI stuff) I would probably not put this in the same workflow as the test and coverage jobs. I would either keep it completely separate like it is or adapt the existing `format.yml` workflow to include all linting related jobs. There isn't a huge difference between these options. Some tooling (like external badge services) make it a little easier to get the status of a workflow rather than a single job in a milti-job workflow, but it isn't a big deal.

.github/workflows/clippy.yaml
alerque: Yes, definitely. This needs to use the same Rust setup tooling as the other workflows. If there is a reason to change that (which their might be but that should be weighed in a different issue/PR) we should change all of them to be consistent.

.github/workflows/clippy.yaml
alerque: I confirmed it is already checking all the members (except the one that is currently excluded and known-broken, that's a matter for a different issue).

.github/workflows/clippy.yaml
pitbuster: changed it to use the same action as the other workflows

waywardmonkeys (approved):

PR #348, opened 2024-03-09
Document current MSRV, note thoolchain, normalize metadata across workspace

alerque:
The MSRV information is the status quo not a change, only documented for tooling that recognizes this field.

Likewise the toolchain file is just for the aid of CI and developer tooling that happens to support it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,15 @@ members = [
 exclude = ["fluent-cli"]
 
 [workspace.package]
+homepage = "https://www.projectfluent.org"
+repository = "https://github.com/projectfluent/fluent-rs"
+license = "Apache-2.0 OR MIT"
+authors = [
+    "Zibi Braniecki <zibi@unicode.org>",
+    "Staś Małolepszy <stas@mozilla.com>"
+]
+categories = ["localization", "internationalization"]
+keywords = ["localization", "l10n", "i18n", "intl", "internationalization"]
 edition = "2021"
 rust-version = "1.67.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,10 @@ members = [
 
 exclude = ["fluent-cli"]
 
+[workspace.package]
+edition = "2021"
+rust-version = "1.67.0"
+
 [workspace.dependencies]
 criterion = "0.5"
 fluent-langneg = "0.13"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,15 +11,17 @@ members = [
   "intl-memoizer",
 ]
 
-exclude = ["fluent-cli"]
+exclude = [
+  "fluent-cli",
+]
 
 [workspace.package]
 homepage = "https://www.projectfluent.org"
 repository = "https://github.com/projectfluent/fluent-rs"
 license = "Apache-2.0 OR MIT"
 authors = [
-    "Zibi Braniecki <zibi@unicode.org>",
-    "Staś Małolepszy <stas@mozilla.com>"
+  "Zibi Braniecki <zibi@unicode.org>",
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -15,6 +15,10 @@ exclude = [
     "fluent-cli",
 ]
 
+[workspace.package]
+edition = "2021"
+rust-version = "1.67.0"
```
waywardmonkeys, line 20: Where is this status quo from?

fluent-cli/rust-toolchain
```diff
@@ -0,0 +1 @@
+stable
```
waywardmonkeys, line 1: Why is this in a subdirectory and not the root? What things actually use this? (I would rather it be a separate PR.)

fluent-pseudo/Cargo.toml
```diff
@@ -4,7 +4,8 @@ description = """
 Pseudolocalization transformation API for use with Project Fluent API.
 """
 version = "0.3.1"
-edition = "2021"
+edition.workspace = true
+rust-version = "1.64.0"
```
waywardmonkeys, line 8: Why is this one and another different? (Or more simply, where is this derived from?)

waywardmonkeys (changes requested):
A couple of questions, but also, I think this would be better as a single PR that starts to use the workspace info, and then something setting up rust-version. It isn’t clear to me why set edition via the workspace, but not license or other info that is the same.

Cargo.toml
alerque: From bisecting builds until I figured out what this can actually be built with, using `cargo msrv`.

fluent-cli/rust-toolchain
alerque: Because it's a mistake. Also `fluent-cli` is quite broken in several ways and trying to work around that while still testing other things was making for a tap-dance ... and I tripped.

fluent-pseudo/Cargo.toml
alerque: Because it *is* different, and I didn't want to artificially boost the MSRV until we decide to do so for some reason, I was only trying to document the status quo. Again this is determined through trial and error with `cargo msrv` as I was feeling my way around where things are at now.

alerque:
Yes I'd be find with splitting this up ... or appending a few more things that can be normalized via workspace meta data. I was just feeling my way around what is currently here so I have a better understanding of what will/won't break when other adjustments are made (e.g. will bumping dependencies as in other PR's change the MSRV).

But since I suggest not landing *anything* at all until we figure out a safe harbor release (which means access to crates.io etc.) I'll leave it together for now for discussion and if we want to refactor it later when we're ready to move on we can of course. I'm comfortable with Git rebase/revise/absorb etc.

fluent-cli/rust-toolchain
alerque: Corrected.

fluent-pseudo/Cargo.toml
alerque: This *probably* doesn't matter since the two primary dependents (`fluent` and `l10n`) both have higer MSRV's than this anyway (the latter being incorrectly documented). Of course that doesn't mean *nobody* else directly depends on this without also depending on other parts of Fluent; but chances are I'd be up for normalizing this in the next (non-safe-harbor) release so the whole workspace specifies the MSVR as the highest minimum of all the members.

waywardmonkeys (changes requested):
Everything seems fine to me for the workspace meta data, but I don't think we need the `rust-toolchain`

We'll need to fix up CI separately to do something more similar to what the Linebender repos do in that regard.

waywardmonkeys (approved):

PR #349, opened 2024-03-15
chore: Release 'safe harbor' versions of Fluent workspace crates

### LINKED ISSUE #347: Does a 'safe harbor' release of the current HEAD make sense before further changes?
Per https://github.com/projectfluent/fluent/issues/358#issuecomment-1986817388, myself and @waywardmonkeys have recently been given permission to further development on this project. Many thanks to those inside Mozilla pursuing this outcome.

I'm opening this issue to invite discussion on the current release cycle. Frequently in the past when I have started maintaining a project that hasn't seen a release in a while, I'll cut a 'safe harbor' release that represents the current state of the repo with only minimal changes needed to make it releasable (e.g. changelog, etc.). This way anybody with concerns about an influx of code being stewarded under different management has an easy place to either pin their dependencies or audit the kind of changes that are being made. My question is does that make sense here.

There are not a *lot * of changes since the last tagged releases, but there are new features, bug fixes, and some minor refactoring that has landed under the guidance of the previous team. I haven't dug into them all yet to even know if this would need to be a semver patch or minor release if we were to make one, and whether the changes made to date are in a shippable state. I'll be looking at that, but if folks have input that would be great.

Next up after determining whether to pursue an immediate release of the *current* state will be reviewing the backlog of PRs and sorting out the state of existing contributions. Having been watching the PRs accumulate I know there is at least some good stuff in there! Then we'll start gearing up for future release cycles!

Thoughts?

alerque:
These releases serve as a markers in the transition from the Project Fluent related crates being a wholly Mozilla managed project to bringing on non-Mozilla community maintainers. The safe harbor release series publishes the current Git HEAD of all crates as a convenience to downstream consumers that either audit the code and contributors, don't want to adopt potential changes for whatever reason, or just want an easy reference point against which to compare future releases.

Closes #347

----

Related issues:

* Background to bringing on non-Mozilla maintainers:
  https://github.com/projectfluent/fluent/issues/358

* Matching PR for fluent-langneg that is in a different workspace:
  https://github.com/projectfluent/fluent-langneg-rs/pull/27

### DIFF
--- a/fluent-bundle/CHANGELOG.md
+++ b/fluent-bundle/CHANGELOG.md
@@ -4,6 +4,21 @@
 
   - …
 
+## fluent-bundle 0.15.3 (March 16, 2024)
+  - This is a 'safe harbor' release prior to bringing on non-Mozilla community maintainers
+  - Implement `From<&String>` for `FluentValue`
+  - Add `FluentValue.into_string` to prevent String clone
+  - Fix `FluentValue::try_number` accepting numbers
+  - Allow optional arguments on `FluentValue`
+  - Fix behavior of `FluentArgs::set`
+  - Resolve function instead in `impl ResolveValue`
+  - Add type alias for concurrent `FluentBundle`
+  - Fix `FluentBundle::format_pattern` lifetimes
+  - Workspace: Update to Rust 2021
+  - Workspace: Add various missing documentation and fix typos and links
+  - Workspace: Cleanup meta-data using workspaces, use SPDX licenses, etc.
+  - Workspace: Apply rustfmt and clippy lint fixes
+
 ## fluent-bundle 0.15.2 (October 25, 2021)
   - Bump `self_cell` to 0.10.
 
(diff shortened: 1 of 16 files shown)


alerque:
@zbraniecki Similar to [my commend on the matching fluent-langneg PR](https://github.com/projectfluent/fluent-langneg-rs/pull/27#issuecomment-1999127702) I think it would make a lot of sense for you to do some combination of ammending this commit to be the author, adding a signoff, and/or doing the final publish to crates.io. Your name is already on many/most of the releases to date so having it there again seems like the clearest signal to give end users clear expectations.

Likewise I set this to draft mode pending any input folks have on whether to add myself and @waywardmorkeys to the authors field as an even more explicit signal an what to expect from future releases.

alerque:
Since nobody seems inclined to comment on the authors issue, my inclination is to just leave it alone and keep the safe-harbor commit as minimal as possible, then change the meta data as soon as we actually start doing something. I'll put it in my pending meta-data cleanup PR (#348).

And with that, I'll take his out of draft mode. I've done some testing and semver checks and everything I've seen suggests this can get merged and published to crates.io.

gregtatum (approved):
This all seems reasonable to me! Thanks for taking the time to do it.

zbraniecki (approved):

PR #351, opened 2024-03-22
Bump criterion to 0.5 to avoid dependency on atty

Pi-Cla:
Criterion before 0.5 depended on atty which is affected by [RUSTSEC-2021-0145](https://rustsec.org/advisories/RUSTSEC-2021-0145)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude = [
 ]
 
 [workspace.dependencies]
-criterion = "0.3"
+criterion = "0.5"
 fluent-langneg = "0.13"
 futures = "0.3"
 iai = "0.1"


Cargo.toml
```diff
@@ -1,22 +1,20 @@
...
 members = [
-    "fluent-syntax",
-    "fluent-bundle",
-    "fluent-fallback",
-    "fluent-resmgr",
-    "fluent-pseudo",
-    "fluent-testing",
-    "fluent",
-    "intl-memoizer"
+  "fluent-syntax",
```
alerque, line 4: Please don't include all the unrelated re-styling in this PR, it is duplicated by #348 anyway.

alerque (changes requested):

Cargo.toml
Pi-Cla: got it, this rebase should address your concerns

alerque (approved):
LGTM, thanks for the update. Now just waiting on #349. As soon as a safe-harbor releases go out I'll test this. Since it's a developer-only dependency and not affecting downstream apps I think it's okay not to rush this.

alerque:
Note to self, merge after #350.

Issue #19, opened 2017-10-02
Implement calling built-in Functions


stasm:
`CallExpressions` currently don't implement the `ResolveValue` trait. In the first iteration let's add a simple `NUMBER()` built-in which doesn't even take any additional arguments.

@zbraniecki What are our options for locale-aware number- and date-formatting?

zbraniecki:
There are none. I have the `fluent-locale` crate almost ready which will add language negotiation and we can mock a simple PluralRules on it. I wanted to use it as the first builtin.

As for number/date formatting, I don't believe there's anything yet in Rust, but I expect things to start popping up soon. For now, we could just add a custom NUMBER function that transliterates to eastern arabic numerals if the language is `ar`.

zbraniecki:
@stasm - I think this is already fixed - you can add custom functions and they can be called. We still await a decent implementation of either `NUMBER` or `DATETIME` to provide built-ins.

Do you want to keep this issue open until we find and implement those two?

zbraniecki:
closing as per lack of activity. reopen if needed.

Xiretza:
I think this should be reopened, the built-in functions are still not implemented, which is either a bug in `fluent-rs` or in https://projectfluent.org/fluent/guide/functions.html.

alerque:
@Xiretza This issue has a lot of unrelated discussion about the implementation needed to have functions in the first place. I suggest opening a new issue (or issues) for the specific function(s) that need to be implemented to be a full spec implementation. That will make it easier to track. Thanks.

Issue #313, opened 2023-02-19
Cannot use built-in functions


lynn2910:
Hello, i have this fluent content:
```
server_info_desc =   > 🏷️ **Quelques statistiques :**
  {"** **"}
  `{ NUMBER($emojis) }` emojis
  `{ NUMBER($stickers) }` stickers
  `{ NUMBER($threads) }` threads
  `{ NUMBER($categories) }` categories
```
But, when i run it, the text displayed is:
```
⁨{NUMBER()}⁩ emojis
⁨{NUMBER()}⁩ stickers
⁨{NUMBER()}⁩ threads
⁨{NUMBER()}⁩ categories
```
So, at first, I was thinking that i wasn't using the good names or syntax, but everything work properly on web playground, and, this is how I'm declaring all arguments:
```rs
            let mut fluent_args = FluentArgs::new();
            {
                fluent_args.set("emojis", emojis_nb);
                fluent_args.set("stickers", stickers_nb);
                fluent_args.set("threads", threads);
                fluent_args.set("categories", categories);
            }
```
Because of all that, the only option is that I missed something when formatting the message.
I use this function:
```rs
  pub fn get(&self, lang: &String, id: &str, args: Option<&FluentArgs<'_>>) -> String {
    if let Some(bundle) = self.lib.get(lang) {
      let msg = bundle.get_message(id);
      if let Some(m) = msg {
        if let Some(value) = m.value() {
          let mut err = vec![];
          bundle.format_pattern(value, args, &mut err).to_string()
        } else {
          NO_VALUE.clone().to_string()
        }
      } else {
        NO_MESSAGE.clone().to_string()
      }
    } else {
      NO_LANG.clone().to_string()
    }
  }
```

Anyone know why this isn't working properly ?

kellpossible:
Perhaps they were never implemented as per https://github.com/projectfluent/fluent-rs/issues/19#issuecomment-466160940 ?

alerque:
This is definitely a valid issue, built in types were never implemented so of course they could not be used.

That being said I think I'm going to close this issue because we have more specific ones for tracking each implementation. A `NUMBER()` function is in progress in #353 that can be tracked there. Additionally a specific issue for `DATETIME()` is at #181.

PR #353, opened 2024-04-26
Add NUMBER builtin function

Xiretza:
Followup to #259. Adds the `type` property to `FluentNumberOptions` and makes it available through a new `NUMBER()` built-in fluent function. Built-ins (currently just `NUMBER()`) are added to the bundle using `FluentBundle::add_builtins()`.

Cc #19.

### DIFF
--- a/fluent-bundle/src/types/number.rs
+++ b/fluent-bundle/src/types/number.rs
@@ -8,6 +8,23 @@ use intl_pluralrules::operands::PluralOperands;
 use crate::args::FluentArgs;
 use crate::types::FluentValue;
 
+#[derive(Debug, Default, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum FluentNumberType {
+    #[default]
+    Cardinal,
+    Ordinal,
+}
+
+impl From<&str> for FluentNumberType {
+    fn from(input: &str) -> Self {
+        match input {
+            "cardinal" => Self::Cardinal,
+            "ordinal" => Self::Ordinal,
+            _ => Self::default(),
+        }
+    }
+}
+
 #[derive(Debug, Copy, Clone, Default, Hash, PartialEq, Eq)]
 pub enum FluentNumberStyle {
     #[default]
@@ -48,6 +65,7 @@ impl From<&str> for FluentNumberCurrencyDisplayStyle {
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub struct FluentNumberOptions {
+    pub r#type: FluentNumberType,
     pub style: FluentNumberStyle,
     pub currency: Option<String>,
     pub currency_display: FluentNumberCurrencyDisplayStyle,
@@ -62,6 +80,7 @@ pub struct FluentNumberOptions {
 impl Default for FluentNumberOptions {
     fn default() -> Self {
         Self {
+            r#type: Default::default(),
             style: Default::default(),
             currency: None,
             currency_display: Default::default(),
@@ -79,6 +98,9 @@ impl FluentNumberOptions {
     pub fn merge(&mut self, opts: &FluentArgs) {
         for (key, value) in opts.iter() {
             match (key, value) {
+                ("type", FluentValue::String(n)) => {
+                    self.r#type = n.as_ref().into();
+                }
                 ("style", FluentValue::String(n)) => {
                     self.style = n.as_ref().into();
                 }
--- a/fluent-bundle/src/builtins.rs
+++ b/fluent-bundle/src/builtins.rs
@@ -1,0 +1,14 @@
+use crate::{FluentArgs, FluentValue};
+
+#[allow(non_snake_case)]
+pub fn NUMBER<'a>(positional: &[FluentValue<'a>], named: &FluentArgs) -> FluentValue<'a> {
+    let Some(FluentValue::Number(n)) = positional.first() else {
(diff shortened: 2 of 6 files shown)


alerque:
Will review, but definitely waiting on #347 before we can actually *do* anything about this ;-) Thanks for taking the time to contribute.

fluent-bundle/src/types/number.rs
```diff
@@ -8,6 +8,28 @@ use intl_pluralrules::operands::PluralOperands;
 use crate::args::FluentArgs;
 use crate::types::FluentValue;
 
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum FluentNumberType {
+    Cardinal,
+    Ordinal,
+}
+
+impl std::default::Default for FluentNumberType {
```
waywardmonkeys, line 17: This can be:

```
#[derive(Debug, Copy, Clone, Default, Hash, PartialEq, Eq)]
pub enum FluentNumberType {
    #[default]
    Cardinal,
    Ordinal,
}
```

And then you don't need the `impl std::default::Default`

I've made this change for the other enums here in #356, which will end up probably conflicting with this PR.

waywardmonkeys (changes requested):
Please rebase and change the derivation of `Default` once #356 lands.

Are there other tests that should be added?

alerque:
I was reviewing this and went ahead and rebased to fix the commit messages and also threw in the `#[derive(Default)` cleanup too.

fluent-bundle/src/types/number.rs
```diff
@@ -8,6 +8,23 @@ use intl_pluralrules::operands::PluralOperands;
 use crate::args::FluentArgs;
 use crate::types::FluentValue;
 
+#[derive(Debug, Default, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum FluentNumberType {
```
alerque, line 12: Is there discussion somewhere I missed about adding this type parameter? I don't see this in the Fluent language spec for NUMBER, and it isn't part of `intl.NumberFormat` either. I understand why it is useful but I think we need to keep an eye on how this Rust implementation stacks up against the spec and other implementations. There might have been discussion or other implementations this is coming from I just don't know where that is/was.

waywardmonkeys: https://github.com/projectfluent/python-fluent/blob/bc5b0e52a059944503ff2624b8d0f9c285d81016/fluent.runtime/fluent/runtime/types.py#L32

https://github.com/projectfluent/fluent/blob/fd8f95478e29dda8121da7e275d375eb8dadbcb0/guide/selectors.md?plain=1#L42

Xiretza: I also don't know where it comes from, but it's in the docs and in other implementations; thanks for the links, @waywardmonkeys.

waywardmonkeys: https://github.com/projectfluent/fluent/pull/259 is the pull that added this to the guide.

It was added to the Python implementation here: https://github.com/projectfluent/python-fluent/pull/193 with a note that it wasn't in the spec but was in the guide.

waywardmonkeys:
A couple of notes:

In #259, @gregtatum mentioned using `kind` rather than `r#type` ... not sure if that's relevant here or not.

Also, https://projectfluent.org/fluent/guide/functions.html#number lists a number of other options. Should any of them be supported as part of this? (or a follow-up issue filed to track that they're missing?)

Xiretza:
> In #259, @gregtatum mentioned using `kind` rather than `r#type` ... not sure if that's relevant here or not.

I changed that back because I wanted to avoid using a different term in Rust than is used everywhere else.

> Also, https://projectfluent.org/fluent/guide/functions.html#number lists a number of other options. Should any of them be supported as part of this? (or a follow-up issue filed to track that they're missing?)

Those should already be handled here: https://github.com/projectfluent/fluent-rs/blob/f6168c8101af7dc39dbfbed286d2b6f9e7271fd9/fluent-bundle/src/types/number.rs#L79

alerque (changes requested):
I think I'm on board with this and most discussion is resolved, but I have to agree with this one:

> Are there other tests that should be added?

How is this getting tested?

Xiretza:
There's a doctest, I can add integration tests too if desired.

alerque (approved):
I saw the example in the docstring but I'd missed the output in the test run. I just confirmed it is there and don't know specifically what else, if anything, would need a test.

alerque:
Back to you @waywardmonkeys. Was there something else specific you thought should be tested?

waywardmonkeys:
I'd probably go with something like the Python tests: https://github.com/projectfluent/python-fluent/pull/193/files#diff-30d1a24a3e67774b5d192acb1cd3f532961bae49746db23ab7b7962e7cb8a25d

Xiretza:
I've added an integration test.

waywardmonkeys (approved):

nazar-pc:
I just tried `main` and still getting `Unknown function: NUMBER()`, am I missing something?

Also `maximumFractionDigits` doesn't seem to be supported as described in the docs, am I right?

This is all a bit confusing, especially with https://github.com/projectfluent/fluent-rs/issues/19 and https://github.com/projectfluent/fluent-rs/issues/313 closed as completed.

PR #354, opened 2024-05-03
Fix doc links.

waywardmonkeys:
One link was broken and the other was using the old HTML style links.


alerque (approved):
LGTM, one more for the books as soon as the safe-harbor release gets tagged.

By the way to make changelog generation easier I'd really like to start using and even enforcing [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so that I can use something like `git-cliff` for future release cycles. The conventional spec can be a bit ugly/inflexible/wordy at times but the tooling support is way better than anything else out there. I already had to come through 2 years of commits to hand write the safe harbor release notes and the commit messages are only partially useful. Would you mind prefixing this commit message with `docs: ...`?

alerque:
Since this was a single-commit PR anyway I just squashed-merged it so I could fix the commit message in the process.

PR #356, opened 2024-05-06
refactor: Derive `Default` for enums

waywardmonkeys:
As of Rust 1.62, enums can use `#[derive(Default)]`.

See <https://github.com/rust-lang/rust/pull/94457>.


alerque (approved):
TIL, I can apply this to a few of my own projects ;-)

PR #357, opened 2024-05-06
Fix syntax error in CI run

alerque:
Follow up to #342 which didn't surface the error because actions don't run in PRs until they have been merged to the default branch

### DIFF
--- a/.github/workflows/clippy.yaml
+++ b/.github/workflows/clippy.yaml
@@ -20,4 +20,4 @@ jobs:
           components: rustfmt
 
       - name: Clippy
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
BTW, setting up to use the merge queue would have prevented that.

alerque:
@waywardmonkeys Yes, unfortunately I don't have access to all the repo settings and in particular the branch protection or CI requirements for PRs. I have in mind we probably should ask for it at some point, bit I also thought it might be best to roll with what we have for a little bit so folks can see we're making a good faith effort and are not on a path to making a bungling mess of things.

PR #360, opened 2024-05-09
Revive CLI tooling

alerque:
The CLI crate has been completely blocked/broken for a while. It doesn't have anything essential or even a lot of function, but for folks getting started with Fluent I think it is useful to have something to play with to at least see how an input file gets parsed to an AST without having to build their own app from scratch. I'd like to expand the functionality a little bit to provide a rudimentry message formatter accessible from the CLI as an example/PoC that might also be useful for some testing.

I'm not sure yet what the implications are of having this enabled in the workspace (I wouldn't want people to need to download/build clap unless they explicitly tried to do something with the CLI) so this PR will be draft-mode until at least that gets resolved.


waywardmonkeys (commented):
It is an easier decision to make to fix the build and add that to CI directly and leave the part that modifies the workspace until later.

I think having some examples (and getting ecosystem usage) would be a good way to make it clear to people how to use things.

alerque:
> an easier decision to make to fix the build and add that to CI directly and leave the part that modifies the workspace

Easier decision maybe but not easier to implement. Cargo is stricter now that it was when this was written, and it was written as part of the workspace, but the workspace has moved on leaving this broken and behind. It's easier to update to work again with the workspace than to split it off with independent dependencies and everything else.

Issue #268, opened 2022-08-06
[Question] Make EVAL function (or builtin?)


juliancoffee:
As far as I know in Fluent we can't parametrise arguments. In a sense that this code will be impossible to parse.

```fluent
skeleton = {$case ->
  [genitive] Skeleton
  [possessive] by Skeleton
}
cultist = {$case ->
  [genitive] Cultist
  [possessive] by Cultist
}
shoot = { $victim } has be shoot { $attacker(case: "possessive") }
```

This is obviously not a problem for English, but it is a problem for other languages with grammatical cases, for example, Slavic ones.
(I translated `the bear was killed by another bear` to Polish and got `niedźwiedź został również zabity przez innego niedźwiedzia`. You can see the difference between `niedźwiedź` and `niedźwiedzia`)

So to solve it, I was thinking about adding EVAL function, which would take its first parameter as a term and pass args to it.
```fluent
shoot = { $victim } has been shoot { EVAL($attacker, case: "possessive") }
```

Doing it externally would require allocating the bundle on the heap so we can capture its address in closure. But the reference to the bundle is always here when we call function, as all functions are stored in the bundle, so this may be some built-in.

juliancoffee:
So, I was trying to create a function for EVAL but lifetimes were stronger than me :)
But I was able to add another branch to `ResolveValue` impl (and WriteValue) and it seems to work.
```rust
            Self::FunctionReference {
                id: ast::Identifier { name: "EVAL", ..},
                arguments,
            } => {
                let (resolved_positional_args, resolved_named_args) =
                    scope.get_arguments(Some(arguments));

                let key = if resolved_positional_args.len() == 1 {
                    match &resolved_positional_args[0] {
                        FluentValue::String(s) => s,
                        _ => return FluentValue::Error,
                    }
                } else {
                    return FluentValue::Error
                };

                let msg = if let Some(msg) = scope.bundle.get_message(&key) {
                    msg
                } else {
                    return FluentValue::Error
                };

                let pattern = if let Some(pat) = msg.value() {
                    pat
                } else {
                    return FluentValue::Error
                };

                FluentValue::String(scope.bundle.format_pattern(
                    pattern,
                    Some(&resolved_named_args),
                    &mut Vec::new()
                ).into_owned().into())
            }

```

juliancoffee:
@zbraniecki how feasible would it be to get something like that?

juliancoffee:
[...]

PR #364, opened 2024-06-05
Custom functions with scope access

PPakalns:
Tries to temporary solve complex translation cases mentioned in:
https://github.com/projectfluent/fluent-rs/issues/268
https://github.com/projectfluent/fluent/issues/80

1) Support functions that can access current scope, can work as Message references.
In future we could allow access in custom functions to additional information about current scope (developer arguments, local arguments, current errors)
Message references are highly needed for more complex translation cases with a lot of possible message combinations.

2) Additionally allow using expressions inside named arguments.  
(It looks like by default fluent doesn't allow it [Playground](https://projectfluent.org/play/?s=eyJtZXNzYWdlcyI6IiMgVHJ5IGVkaXRpbmcgdGhlIHRyYW5zbGF0aW9ucyBiZWxvdy5cbiMgU2V0ICR2YXJpYWJsZXMnIHZhbHVlcyBpbiB0aGUgQ29uZmlnIHRhYi5cblxuXG4tc2hhcmVkLXBob3RvcyA9XG4gICAgeyR1c2VyTmFtZX0geyRwaG90b0NvdW50IC0-XG4gICAgICAgIFtvbmVdIGFkZGVkIGEgbmV3IHBob3RvXG4gICAgICAgKltvdGhlcl0gYWRkZWQgeyRwaG90b0NvdW50fSBuZXcgcGhvdG9zXG4gICAgfSB0byB7JHVzZXJHZW5kZXIgLT5cbiAgICAgICAgW21hbGVdIGhpcyBzdHJlYW1cbiAgICAgICAgW2ZlbWFsZV0gaGVyIHN0cmVhbVxuICAgICAgICpbb3RoZXJdIHRoZWlyIHN0cmVhbVxuICAgIH0uXG4gICAgXG5mYWlsID0gMS4geyAtc2hhcmVkLXBob3Rvcyh1c2VyTmFtZTogJHVzZXJOYW1lLCB1c2VyR2VuZGVyOiAkdXNlckdlbmRlcikgfVxub2sgPSAyLiB7IC1zaGFyZWQtcGhvdG9zKHVzZXJOYW1lOiBcInRlc3RcIiwgdXNlckdlbmRlcjogXCJ0ZXN0XCIpIH1cbiIsInZhcmlhYmxlcyI6eyJ1c2VyTmFtZSI6IkFubmUiLCJ1c2VyR2VuZGVyIjoiZmVtYWxlIiwicGhvdG9Db3VudCI6M30sInNldHVwIjp7InZpc2libGUiOlsibWVzc2FnZXMiLCJvdXRwdXQiXSwibG9jYWxlIjoiZW4tVVMiLCJkaXIiOiJsdHIifX0) )
This functionality can be placed under boolean flag if we do not want to enable it by default.

See examples included in added tests.

Would like to get feedback if these changes can be upstreamed to fluent-rs.

### DIFF
--- a/fluent-bundle/src/args.rs
+++ b/fluent-bundle/src/args.rs
@@ -52,7 +52,7 @@ use crate::types::FluentValue;
 ///     "Hello, John. You have 5 messages."
 /// );
 /// ```
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct FluentArgs<'args>(Vec<(Cow<'args, str>, FluentValue<'args>)>);
 
 impl<'args> FluentArgs<'args> {
--- a/fluent-bundle/src/bundle.rs
+++ b/fluent-bundle/src/bundle.rs
@@ -24,6 +24,7 @@ use crate::message::FluentMessage;
 use crate::resolver::{ResolveValue, Scope, WriteValue};
 use crate::resource::FluentResource;
 use crate::types::FluentValue;
+use crate::FluentFunctionObject;
 
 /// A collection of localization messages for a single locale, which are meant
 /// to be used together in a single view, widget or any other UI abstraction.
@@ -535,6 +536,13 @@ impl<R, M> FluentBundle<R, M> {
     pub fn add_function<F>(&mut self, id: &str, func: F) -> Result<(), FluentError>
     where
         F: for<'a> Fn(&[FluentValue<'a>], &FluentArgs) -> FluentValue<'a> + Sync + Send + 'static,
+    {
+        self.add_function_with_scope(id, func)
+    }
+
+    pub fn add_function_with_scope<F>(&mut self, id: &str, func: F) -> Result<(), FluentError>
+    where
+        F: FluentFunctionObject + Sync + Send + 'static,
     {
         match self.entries.entry(id.to_owned()) {
             HashEntry::Vacant(entry) => {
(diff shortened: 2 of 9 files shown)


PPakalns:
Ping.

alerque:
@zbraniecki or @waywardmonkeys (or any other maintainers) has anybody had a spare minute to give this a quick gander?

zbraniecki:
Ehh, some of that is by design (@stasm may remember more on why we didn't want to allow for expr inside named arguments).

Message References are desirable feature, but are non-trivial to implement and I'd prefer to implement them in line with how @eemeli is planning to support them in MF2.0. @eemeli - can you look at the tests in this PR and comment on the alignment?

As for the code - I only skimmed through and I do hope that we can improve the impl before we land this. My recommendation would be to take more time on this and not try to rush it.

eemeli:
This seems pretty well aligned with MF2's custom functions, which is how it allows for something like messages references. I do have some concern about a general-purpose `MSGREF()` being potentially more difficult for translators to work with compared to more specific functions like `HERO()` or `CREATURE()` that can be much more precisely defined.

Also, splitting up a message like the example `hero-owns-creature` into chunks makes it more difficult to translate as a whole, prohibits allowing "owns" to vary depending on the creature type, and quite easily leads to unexpected whitespace in languages like Chinese or Japanese that don't use it. So while the approach is technically sound, there are some potential footguns here that ought to be accounted for.

PR #367, opened 2024-08-03
Remove debug print in `NUMBER` builtin function

nazar-pc:
Looks like it was forgotten by accident in https://github.com/projectfluent/fluent-rs/pull/353/files#diff-0bf32362310a2aa403a7e651a2e8fe562fb15d1301659645b933ec312d2a64f4R11 and results in A LOT of noise in logs.


alerque (approved):
Thanks for the contribution!

PR #370, opened 2024-09-05
Relax patch-level dependency pin on intl_pluralrules

alerque:
The [intl_pluralrules](https://crates.io/crates/intl_pluralrules/versions) dependency was pinned to a patch level release in ba16827, but I can't find any justification for it rematining that way. Since then we'ne unpinned alll remaining 3rd party crates to minor version levels instead of patch levels to give more flexability to downstream builds. We only pin patch realeases for other things in this workspace that we can reasonably expect to match at any given point.

The commit bump to v7.0.1 also happened before the v7.0.2 release existed, and looking at [the upstream commits](https://github.com/zbraniecki/pluralrules/commits/master/) doesn't suggest to me any reason we shouldn't let people start building against that.


waywardmonkeys (approved):
It isn't "pinned" as in "=7.0.1" but just stating that "7.0.1" is the minimum supported. The changes between "7.0.0" and "7.0.1" seem pretty minimal though.

Issue #270, opened 2022-08-17
Expose spans in parser / source positions in the AST


juliancoffee:
I'm thinking about writing my own tool for fluent files and if I want to use something like git blame to find out updates of messages, I'll need to know the span of entry. fluent-js seems to expose that and I guess fluent-rs operates on spans too, as it has it in errors, but I couldn't find it in public API, is it possible to expose it?

zbraniecki:
Yup, we did not include it in Rust parser, but the intent was always to eventually add it for the purpose you're describing.

Please, feel free to prototype - the main condition is that it should not degrade runtime performance. That may mean we need to keep it as a separate AST/parser or via some generics/macros generalize the current AST/parser to allow for white space preservation.

alerque:
The later issue #346 seems to be a duplicate of this. There is actually a bit more discussion there that might be worth reviewing before anybody jumps in to start implementing this.

mondeja:
What about a `cst` (Concrete Syntax Tree) feature for `fluent-syntax` that would add this information to the current AST parser?

PR #373, opened 2024-11-16
Add feature to add spans tracking source positions in AST nodes

Ertanic:
I needed a parser for .ftl files. I found tree-sitter-fluent, but for some reason it couldn't parse a valid file, throwing errors when trying to use replaceable expressions. Decided to use fluent-syntax, but why does the javascript version have node spans but the rust version does not. This PR solves this issue, but since there is usually no need for spans, I hid them behind the `spans` feature.

And also to avoid conflicts in tests, because there the tree is formatted, because of which the spans change, the implementation version of `PartialEq` for AST nodes was divided into a `derive` implementation and a manual one.

The good idea is to write tests to match the spans, but I'm not sure how best to do that, I need help with this.

### DIFF
--- a/fluent-syntax/tests/fixtures/benches/contexts/browser/appmenu.spans.json
+++ b/fluent-syntax/tests/fixtures/benches/contexts/browser/appmenu.spans.json
@@ -1,0 +1,270 @@
+{
+  "body": [
+    {
+      "type": "Comment",
+      "content": [
+        "This Source Code Form is subject to the terms of the Mozilla Public",
+        "License, v. 2.0. If a copy of the MPL was not distributed with this",
+        "file, You can obtain one at http://mozilla.org/MPL/2.0/."
+      ],
+      "span": {
+        "start": 0,
+        "end": 198
+      }
+    },
+    {
+      "type": "GroupComment",
+      "content": [
+        "App Menu"
+      ],
+      "span": {
+        "start": 200,
+        "end": 211
+      }
+    },
+    {
+      "type": "Message",
+      "id": {
+        "name": "appmenuitem-protection-report-title",
+        "span": {
+          "start": 213,
+          "end": 248
+        }
+      },
+      "value": {
+        "elements": [
+          {
+            "type": "TextElement",
+            "value": "Privacy Protections",
+            "span": {
+              "start": 251,
+              "end": 271
+            }
+          }
+        ],
+        "span": {
+          "start": 251,
+          "end": 271
+        }
+      },
+      "attributes": [],
+      "comment": null,
+      "span": {
+        "start": 213,
+        "end": 271
+      }
+    },
+    {
+      "type": "Message",
+      "id": {
+        "name": "appmenuitem-protection-report-tooltip",
+        "span": {
+          "start": 271,
+          "end": 308
+        }
+      },
+      "value": null,
+      "attributes": [
+        {
+          "id": {
+            "name": "tooltiptext",
+            "span": {
+              "start": 316,
+              "end": 327
+            }
+          },
+          "value": {
+            "elements": [
+              {
+                "type": "TextElement",
+                "value": "Go to your privacy report",
+                "span": {
+                  "start": 330,
+                  "end": 356
+                }
+              }
+            ],
+            "span": {
+              "start": 330,
+              "end": 357
+            }
+          },
+          "span": {
+            "start": 356,
+            "end": 357
+          }
+        }
+      ],
+      "comment": null,
+      "span": {
+        "start": 271,
+        "end": 357
+      }
+    },
+    {
+      "type": "GroupComment",
+      "content": [
+        "Zoom Controls"
+      ],
+      "span": {
+        "start": 357,
+        "end": 373
+      }
+    },
+    {
+      "type": "Message",
+      "id": {
+        "name": "appmenuitem-zoom-enlarge",
+        "span": {
+          "start": 375,
+          "end": 399
+        }
+      },
+      "value": null,
+      "attributes": [
+        {
+          "id": {
+            "name": "label",
+            "span": {
+              "start": 405,
+              "end": 410
+            }
+          },
+          "value": {
+            "elements": [
+              {
+                "type": "TextElement",
+                "value": "Zoom in",
+                "span": {
+                  "start": 413,
+                  "end": 421
+                }
+              }
+            ],
+            "span": {
+              "start": 413,
+              "end": 421
+            }
+          },
+          "span": {
+            "start": 420,
+            "end": 421
+          }
+        }
+      ],
+      "comment": null,
+      "span": {
+        "start": 375,
+        "end": 421
+      }
+    },
+    {
+      "type": "Message",
+      "id": {
+        "name": "appmenuitem-zoom-reduce",
+        "span": {
+          "start": 421,
+          "end": 444
+        }
+      },
+      "value": null,
+      "attributes": [
+        {
+          "id": {
+            "name": "label",
+            "span": {
+              "start": 450,
+              "end": 455
+            }
+          },
+          "value": {
+            "elements": [
+              {
+                "type": "TextElement",
+                "value": "Zoom out",
+                "span": {
+                  "start": 458,
+                  "end": 467
+                }
+              }
+            ],
+            "span": {
+              "start": 458,
+              "end": 468
+            }
+          },
+          "span": {
+            "start": 467,
+            "end": 468
+          }
+        }
+      ],
+      "comment": null,
+      "span": {
+        "start": 421,
+        "end": 468
+      }
+    },
+    {
+      "type": "GroupComment",
+      "content": [
+        "Firefox Account toolbar button and Sync panel in App menu."
+      ],
+      "span": {
+        "start": 468,
+        "end": 529
+      }
+    },
+    {
+      "type": "Message",
+      "id": {
+        "name": "fxa-toolbar-sync-now",
+        "span": {
+          "start": 531,
+          "end": 551
+        }
+      },
+      "value": null,
+      "attributes": [
+        {
+          "id": {
+            "name": "label",
+            "span": {
+              "start": 559,
+              "end": 564
+            }
+          },
+          "value": {
+            "elements": [
+              {
+                "type": "TextElement",
+                "value": "Sync Now",
+                "span": {
+                  "start": 567,
+                  "end": 576
+                }
+              }
+            ],
+            "span": {
+              "start": 567,
+              "end": 576
+            }
+          },
+          "span": {
+            "start": 575,
+            "end": 576
+          }
+        }
+      ],
+      "comment": null,
+      "span": {
+        "start": 531,
+        "end": 576
+      }
+    }
+  ],
+  "span": {
+    "start": 0,
+    "end": 576
+  }
+}
--- a/fluent-syntax/tests/fixtures/benches/contexts/browser/brand.spans.json
+++ b/fluent-syntax/tests/fixtures/benches/contexts/browser/brand.spans.json
@@ -1,0 +1,239 @@
+{
+  "body": [
+    {
+      "type": "Comment",
+      "content": [
+        "This Source Code Form is subject to the terms of the Mozilla Public",
+        "License, v. 2.0. If a copy of the MPL was not distributed with this",
+        "file, You can obtain one at http://mozilla.org/MPL/2.0/."
+      ],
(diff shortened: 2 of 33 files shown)


alerque:
This would address #270, no?

Have you run any benchmarks with/without this feature enabled?

fluent-syntax/src/serializer.rs
```diff
@@ -478,12 +484,24 @@ mod test {
     macro_rules! text_message {
         ($name:expr, $value:expr) => {
             Entry::Message(Message {
-                id: Identifier { name: $name },
+                id: Identifier {
+                    name: $name,
+                    #[cfg(feature = "spans")]
+                    span: Span::new(0..0),
```
zbraniecki, line 490: can you use `Span::default()` here?

fluent-syntax/src/serializer.rs
```diff
@@ -478,12 +484,24 @@ mod test {
...
+                        #[cfg(feature = "spans")]
+                        span: Span::new(0..0),
+                    }],
+                    #[cfg(feature = "spans")]
+                    span: Span::new(0..0),
                 }),
                 attributes: vec![],
                 comment: None,
+                #[cfg(feature = "spans")]
+                span: Span::new(0..0),
```
zbraniecki, line 504: can you use `Span::default()` here and above?

fluent-syntax/src/serializer.rs
```diff
@@ -563,14 +581,34 @@ mod test {
...
+                            #[cfg(feature = "spans")]
+                            Span::new(0..0),
+                        ),
+                        #[cfg(feature = "spans")]
+                        span: Span::new(0..0),
+                    },
+                    PatternElement::TextElement {
+                        value: " bar",
+                        #[cfg(feature = "spans")]
+                        span: Span::new(0..0),
```
zbraniecki, line 603: can you use `Span::default()` here and above?

fluent-syntax/src/serializer.rs
```diff
@@ -630,6 +668,8 @@ mod test {
         let mut ast = parse("foo = bar\n").expect("failed to parse ftl resource");
         ast.body[0].as_message().comment.replace(Comment {
             content: vec!["great message!"],
+            #[cfg(feature = "spans")]
+            span: Span::new(0..0),
```
zbraniecki, line 672: can you use `Span::default()` here?

fluent-syntax/src/serializer.rs
```diff
@@ -563,14 +581,34 @@ mod test {
...
+                        span: Span::new(0..0),
                     },
-                    PatternElement::TextElement { value: " bar" },
                 ],
+                #[cfg(feature = "spans")]
+                span: Span::new(0..0),
             },
             default: false,
+            #[cfg(feature = "spans")]
+            span: Span::new(0..0),
```
zbraniecki, line 611: can you use `Span::default()` here?

zbraniecki (commented):
all in all, happy to see that! I definitely was hoping we'll gain this feature one day. Thank you for contributing!

My only uber-concern is that you use PartialEq to compare elements with different span. I'm not sure how canonical it is to use PartialEq this way. We wrangled with the meaning and role of PartialEq in ICU4X for a long time and I'm still not sure how to handle it but in ICU4X we decided to stay on the cautious side and introduce a function like `cmp_value` to allow for comparisons that exclude part of the value.
Span can be thought of as a metadata or part of the element and with the trait use you make it impossible to compare *with spans*.

I'm not sure what's the right way around it and if canonically ASTs in Rust (or other languages) use, and it is not uncommon to compare skipping spans, I'm fine with doing the same.

zbraniecki:
> And I don't quite understand your point. Are you proposing to introduce additional methods for fields to compare structures and their fields? Or to compare all fields of node structures separately from PartialEq and Eq in separate methods?

I'm raising a concern that semantically the following code should pass:

```
let node1 = Node {
  value: "foo",
  span: span!(0, 4),
};

let node2 = Node {
  value: "foo",
  span: span!(5, 11),
};
assert_ne!(node1, node2);
```

because those two nodes are *not* equal. Their content is different.

Now, what is true is that in *most* cases we care about the actual content of the node, not its meta information. We can explicitly achieve that by doing:

```
let node1 = Node {
  value: "foo",
  span: span!(0, 4),
};

let node2 = Node {
  value: "foo",
  span: span!(5, 11),
};
assert_ne!(node1, node2);

// Option 1:
assert_eq!(node1.content, node2.content);

// Option 2:
assert_eq!(node1.cmp_content(&node2));
```

Or we can do what is proposed in the PR and add:

```
let node1 = Node {
  value: "foo",
  span: span!(0, 4),
};

let node2 = Node {
  value: "foo",
  span: span!(5, 11),
};
assert_eq!(node1, node2);

assert_ne!(node1.span, node2.span);
assert_eq!(node1.cmp_span(&node2));
```

I'm not sure what is the most common approach to AST comparisons with spans. I'd suggest checking prior art in other parser/AST/serializer models.

Ertanic:
Okay...

fluent-syntax/src/ast/mod.rs
```diff
@@ -1448,12 +1520,50 @@ pub enum Expression<S> {
...
+/// assert_eq!(resource.span, Span { start: 0, end: 27 });
+/// assert_eq!(id.span, Span { start: 0, end: 11 }); // the span of hello-world identifier
+/// ```
+#[cfg(feature = "spans")]
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Span {
+    pub start: usize,
+    pub end: usize,
+}
```
zbraniecki, line 1559: did we discuss the value of having this struct be a range itself? It seems like the shape is exactly `Range`, and if we were to use `Range` here we get all of stdlib features of it. Why not?

zbraniecki (commented):
@alerque feel free to decide if you want to merge or agree that we should use `Range` here.
The rest looks good to me! Thank you @Ertanic!

fluent-syntax/src/ast/mod.rs
Ertanic: @zbraniecki, I replaced the Span structure with std::ops::Range<usize>, take a look, please

zbraniecki: This looks good to me! Thank you!
One change I'd likely make is to use a newtype Span(Range<>) with deref to Range. This makes it easy for us to control the validation and impl.

fluent-syntax/Cargo.toml
```diff
@@ -37,7 +37,8 @@ serde_json.workspace = true
 glob = "0.3"
 
 [features]
-default = []
+default = ["spans"]
```
zbraniecki, line 40: I would not enable this by default. I think it's is an optional feature needed for editors/LSPs.

zbraniecki (commented):
This looks good to me, except that I wouldn't suggest enabling spans by default - the default use is to parse for runtime use of the AST, not for editor cycle.

I also like the wrapping of Range because it allows us to swap for the new Range type once it stabilizes and enable Copy - https://hackmd.io/@uhs6rVdLTSS0gnie4q0fqA/ryjYJW2pa

Ertanic:
I'm aware of the feature, I fixed it a long time ago, but the commit hung
image

Ertanic:
I don't know why github can't output my commit, but fork has it: https://github.com/projectfluent/fluent-rs/commit/460c36514d5948172766f396f8a0214be2c6dd1e

Ertanic:
Had to play around with reverting to fix the github. @zbraniecki

alerque:
Is there a Git attribute we need to set to make checking out the new fixtures that explicitly test carriage returns from getting smudged on checkout?

Ertanic:
I couldn't think of anything better than just disabling the documentation tests. Otherwise it's too much work to modify the documentation, which I think will only make it harder to read. From the suggestions to create a separate task for testing documentation, skipping the spans feature.

fluent-syntax/src/ast/helper.rs
```diff
@@ -1,25 +1,49 @@
...
+use super::Span;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-use super::Comment;
 // This is a helper struct used to properly deserialize referential
 // JSON comments which are single continuous String, into a vec of
 // content slices.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
alerque, line 10: Was the removal of `Eq` from derived traits intentional (here an elsewhere)?

fluent-syntax/Cargo.toml
```diff
@@ -63,3 +64,6 @@ required-features = ["json"]
 name = "parser_fixtures"
 path = "tests/parser_fixtures.rs"
 required-features = ["json"]
+
+[lib]
+doctest = false
```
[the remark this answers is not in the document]
Ertanic: Because the test task is started with the `--all-features` flag.

Ertanic:
I've tried fixing the error during the `parse_fixtures_compare` test, even adding spans to the json files, but it still keeps generating the error of no span field in the json file.

alerque:
Please don't do any merging into this branch. I've force pushed several times **to clean up the history** and you keep pushing the dirty commits back into the timeline. Do feel free to work on it if there is something to do, but before committing either `git pull --rebase` to work from where this branch is (or `git reset --hard <COMMIT>` to the lastest commit you see here) and then just commit the changes on top of that. I can selectively squash fixes into the appropriate previous commits much easier than unraveling the cruft that comes in when you merge from your own outdated history. Thanks.

fluent-syntax/tests/serializer_fixtures.rs
```diff
@@ -57,15 +45,33 @@ fn roundtrip_unnormalized_fixtures() {
...
+        )
+        .unwrap();
+        assert_eq!(formatted_content, reserialized);
+        let formatted = parse(formatted_content.as_str()).unwrap_or_else(|(res, _)| res);
+
+        let formatted_content_without_junk =
+            fs::read_to_string(formatted_path.join(path.file_name().unwrap())).unwrap();
+        assert_eq!(formatted_content_without_junk, reserialized_without_junk);
+        let formatted_without_junk =
+            parse(formatted_content_without_junk.as_str()).unwrap_or_else(|(res, _)| res);
```
alerque, line 68: The more I fiddle with it trying to figure out what's going on with tests the more I think this approach to having expanded files for just the serialized message is not the right approach to testing. We should be able to test the original JSON fixtures and just use a query engine of some kind to extract the message bit and work around having serialized position data in our actual results.

With all these expanded and multiple entries in different formats in becomes much harder to audit what tests are actually doing and compare with other implementations.

alerque (changes requested):
Is there anywhere here the span *results* are actually being tested? Mostly I see the tests modified to try to dodge the span info, but we could use at least a couple inputs with the expected position data tested.

See my other comment about the pre-formated test results. I'm a bit skeptical about that being the most maintainable solution, especially since we can't even work out the line ending handling on the file reads. Having the expectation in JSON (as much as I actually despise JSON) does have the advantage of being unambiguous about what the EOL/EOF business is. How would those be updated? Can we backtrack and just use the JSON expectations and figure out how to query the relevant bits we need out of the upstream fixture expectations vs. our serialized data?

alerque:
I was pretty excited to get this feature landed before cutting a release, but I think I'm going to have to delay it. Hopefully we don't have to delay it long—I have no objection to doing a new release on a relatively short timeline as soon as this is actually ready.

My concerns are mostly related to testing, but I was playing around with fixing the tests and it feels like something is wrong with the scruct itself. Enabling the feature (or not) should not immediately break all existing usage of the serialization. That would make it very hard for existing systems with serialized data to migrate. If possible I would really like existing apps not to break the moment somebody tries enabling the feature. That may not be quite possible, but we should at least think through what the ramifications are and document how/why a change needs to be made to keep using existing code easily on the new version.

Until that gets thought through and tests pass both with and without the feature enabled I can't reasonably merge this, and since we have problems with old dependencies starting to hinder usage I kind of need to get an update out to address that.

Again I *do* want this feature to land and am willing to facilitate a release cycle special for it as soon as we're actually comfortable with the upgrade path if there are any breaking changes and tests work well enough to rely on.

Ertanic:
I do not know how to fix the `roundtrip_unnormalized_fixtures` test, so I suggest hiding it for now behind the `cfg(not(feature = "spans"))` attribute?

PR #375, opened 2024-11-29
Apply new clippy lint fixes

alerque:
The first commit here has some obvious fixes for convoluted code that has idiomatic Rust alternatives.

The second commit is for using lifetime elision instead of annotations. I understand this one is a bit more controversion and some projects have opted to ignore this lint instead. Personally I think I'm fine with it, I'd rather see annotations where they are meaningful and need to be included in reading the code to grok it, but if anybody here has issues with that I'm happy to consider ignoring these.

### DIFF
--- a/fluent-bundle/examples/custom_type.rs
+++ b/fluent-bundle/examples/custom_type.rs
@@ -33,7 +33,7 @@ enum DateTimeStyleValue {
 
 // This is just a helper to make it easier to convert
 // a value provided to FluentArgs into an option value.
-impl<'l> From<&FluentValue<'l>> for DateTimeStyleValue {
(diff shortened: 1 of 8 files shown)


alerque:
@waywardmonkeys Any idea off the top of your head how to deal with the other new lint not covered here yet involving mutable references to mutable static?

waywardmonkeys:
> @waywardmonkeys Any idea off the top of your head how to deal with the other new lint not covered here yet involving mutable references to mutable static?

Not offhand and don't have the time this week to dig into it. :(

waywardmonkeys (approved):
These changes so far look good to me.

PR #382, opened 2025-05-19
Bump deps and prep for release

### LINKED ISSUE #381: Please release a new version of fluent-bundle
The last update has been a year ago and it depends on some old libs like self_cell (which is now 4 years ago)
https://crates.io/crates/fluent-bundle/0.15.3/dependencies
https://github.com/Voultapher/self_cell/releases

Same story for rust-hash

It makes cargo-deny sad for some packages:
```

error[duplicate]: found 2 duplicate entries for crate 'self_cell'
    ┌─ /github/workspace/Cargo.lock:225:1
    │  
225 │ ╭ self_cell 0.10.3 registry+https://github.com/rust-lang/crates.io-index
226 │ │ self_cell 1.2.0 registry+https://github.com/rust-lang/crates.io-index
    │ ╰─────────────────────────────────────────────────────────────────────┘ lock entries

```

```

error[duplicate]: found 2 duplicate entries for crate 'rustc-hash'
    ┌─ /github/workspace/Cargo.lock:217:1
    │  
217 │ ╭ rustc-hash 1.1.0 registry+https://github.com/rust-lang/crates.io-index
218 │ │ rustc-hash 2.1.1 registry+https://github.com/rust-lang/crates.io-index
    │ ╰──────────────────────────────────────────────────────────────────────┘ lock entries
```

Thanks!

alerque:
Instigated by #381, but *not* yet adding new release versions since I'm reviewing the possibility of landing #373 first.

### DIFF
--- a/fluent-pseudo/src/lib.rs
+++ b/fluent-pseudo/src/lib.rs
@@ -1,3 +1,4 @@
+use once_cell::sync::Lazy;
 use regex::Captures;
 use regex::Regex;
 use std::borrow::Cow;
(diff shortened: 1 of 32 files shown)


fluent-pseudo/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+use once_cell::sync::Lazy;
```
alerque, line 1: @zbraniecki / @waywardmonkeys Any quick opinions about whether it makes sense given its position only in fluent-pseudo to introduce this dependency to get rid of the reliance on unsafe undefined behavior? If/when we bump the MSRV up a bit we could of course get rid of this again and use the implementation of this in *std*, but this way we can at least do a minor release without affecting our current MSRV while also not having to suppress clippy's objections.

zbraniecki: this looks good to me.

zbraniecki (approved):

PR #383, opened 2025-05-21
Release all affected crates in workspace

alerque:
Closes #381

Follow up on #382 that I merged early *before* these version bumps to get CI tests and lints working as expected while we touch up #373.

PR #386, opened 2025-05-22
Use breaking version bump for fluent

### LINKED ISSUE #384: fluent 0.16.2 contains breaking changes
`fluent` 0.16.2 changed its `fluent-bundle` dependency from 0.15.3 to 0.16. `fluent-bundle` 0.16.0 changed its `fluent-syntax` dependency from 0.11.1 to 0.12.

`fluent` has types from `fluent-syntax` (at least `ast::Pattern` and `ast::Entry`) in its public API, so that `fluent` 0.16.1 uses types from `fluent-syntax` 0.11 while `fluent` 0.16.2 uses types from `fluent-syntax` 0.12. This is a breaking change as they are technically different types.

The issue manifests at least in `i18n-embed` (0.15.4) as it uses `fluent` 0.16 with types from `fluent-syntax` 0.11. This works fine with 0.16.1 (and I suppose 0.16.0),  but breaks with 0.16.2 which expects `fluent-syntax` 0.12 types.

Reproducer:
```sh
cargo init
cargo add i18n-embed -F fluent-system
cargo build
💥
```

### LINKED ISSUE #385: It seems `fluent_bundle` is a "public dependency"  of `fluent`. So a semver bump in `fluent_bundle` should require a semver bump in `fluent`.
> It seems `fluent_bundle` is a "public dependency"  of `fluent`. So a semver bump in `fluent_bundle` should require a semver bump in `fluent`.
>
> The result of this is that projects depending on `fluent` and `i18n-embed` are getting errors about incompatible types (from different versions of `fluent-syntax`) when running `cargo update`.

_Originally posted by @ids1024 in https://github.com/projectfluent/fluent-rs/issues/383#issuecomment-2902814140_

alerque:
* Closes #384
* Closes #385
* See also comments in #383.

It seems we have an issue with downstream projects using (indirectly) `fluent-bundle::*` while at the same time the upstream project is using `fluent::fluent_bundle::*`. This is kind of a weird arrangement, but given that we're exposing it it does seem like the resolver has trouble getting the right combination.

Note `cargo semver-checks` did *not* catch this, but that may be a bug.

If we yank 0.16.2 and replace it with 0.17.0 we might mitigate some of the pain downstream.

Thoughts?

### DIFF
--- a/fluent/CHANGELOG.md
+++ b/fluent/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
-## fluent 0.16.2 (May 20, 2025)
+## fluent 0.17.0 (May 23, 2025)
+  - Bump version to reflect breaking changes in re-exported FluentBundle
+
+## fluent 0.16.2 (May 20, 2025) [yanked]
   - Cleanup docs
 
 ## fluent 0.16.1 (March 16, 2024)
--- a/fluent/Cargo.toml
+++ b/fluent/Cargo.toml
@@ -4,7 +4,7 @@ description = """
 An umbrella crate exposing the combined features of fluent-rs crates with additional convenience macros for Project Fluent,
 a localization system designed to unleash the entire expressive power of natural language translations.
 """
-version = "0.16.2"
+version = "0.17.0"
 edition.workspace = true
 rust-version.workspace = true
 homepage.workspace = true


alerque:
I just yanked fluent-0.16.2.

zbraniecki (approved):
thanks for responding so fast!

Issue #374, opened 2024-11-25
(De)serialize FluentArgs


TheCataliasTNT2k:
Is there a way, to store `FluentArgs` in a database or file?
If not, could we get an implementation for serde Serialize and serde Deserialize?

This would allow said feature to create a non translated string and translate it later.

alerque:
Off the top of my head, I don't think we'd want to saddle everybody with another dependency for this, although it *might* be reasonable to provide such a trait as an optional non-default feature.

In the mean time I can't help but wonder why you don't just serialize your data *before* you stuff it into `FluentArgs` instead of afterwards. Wouldn't that make more sense anyway?

TheCataliasTNT2k:
The idea is not the store a translated text, but the prepared translation itself.

I want to be able to prepare a text, which could be translated. Later, I want to load this text back into memory and translate it into any language I want (if it exists).

Sure, I could create a custom Container, which just maps &str on &str like FluentArgs does. 
When translating, I could convert this container to FluentArgs just as if I only used that.
I will do it this way for now, but I thought, there is already a container for it, so why not use that one?

alerque:
Serde isn't yet a run time dependency for `fluent-bundle`, so this would have to be an optional non-default feature.

I still can't make out what the use case is really. If you have data why can't you just keep it? Or keep a translation. I don't understand what the benefit of storing an intermediary format would be.

PR #388, opened 2025-05-23
Make `Serializer` and some of his methods public

andrew000:
I use this methods to serialize text into Message/Term/etc in my Fluent Extractor.

It would be great to make them public

### DIFF
--- a/fluent-syntax/src/serializer.rs
+++ b/fluent-syntax/src/serializer.rs
@@ -64,14 +64,14 @@ pub fn serialize_with_options<'s, S: Slice<'s>>(
 }
 
 #[derive(Debug)]
-struct Serializer {
+pub struct Serializer {
     writer: TextWriter,
     options: Options,
     state: State,
 }
 
 impl Serializer {
-    fn new(options: Options) -> Self {
+    pub fn new(options: Options) -> Self {
         Serializer {
             writer: TextWriter::default(),
             options,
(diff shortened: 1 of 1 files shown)


alerque:
As the original movers and drivers of the serializer feature getting merged at all @Michael-F-Bryan and @RumovZ, do you have any input relevant to making these bits public?

alerque:
Tangentially related, #374.

RumovZ:
Nope, wouldn't mind.

andrew000:
Any news?

andrew000:
Are we ready to merge? @alerque

andrew000:
Any news?

andrew000:
Any news?

danielrainer:
We're currently looking into building tooling around converting gettext PO files to FTL files and adding some utilities for renaming identifiers and keeping FTL files tidy for fish[^0]. For this, we would need to be able to parse FTL files, modify the parsed data, and serialize it back so that we can update the FTL files. I just started looking into the available APIs from `fluent-rs`, but having a serializer available is almost certainly something we would need.

This PR is very small (it only adds a few `pub` modifiers) and has already been approved. Is there anything preventing it from being merged?

[^0]: https://github.com/fish-shell/fish-shell/pull/11928

waywardmonkeys:
Since this was approved, I'm going to go ahead and land this.

PR #394, opened 2025-11-10
feat: Make `fluent_syntax::parser::core::Parser` pub

danielrainer:
This allows creating a simple FTL formatter by parsing and then immediately serializing the parsing result again.


danielrainer:
I just played around with this some more and it turns out that with `Parser` accessible, quite a lot becomes possible. For example, I just built a POC tool for renaming IDs and variables on top of `fluent-syntax`.

